PR #177, opened 2014-02-03
SES breaks implicit views in the root.

bdw429s:
This is my proposed fix for https://ortussolutions.atlassian.net/browse/COLDBOX-245 but I don't know if the code I removed was there for a reason.

### DIFF
--- a/system/interceptors/SES.cfc
+++ b/system/interceptors/SES.cfc
@@ -125,15 +125,11 @@ Description :
 					routedStruct[ key ] = aRoute[ key ];
 				}
 			}
-
+				
 			// Create Event To Dispatch if handler key exists
 			if( structKeyExists( aRoute,"handler" ) ){
-				// If no action found, default to the convention of the framework, must likely 'index'
-				if( NOT structKeyExists(aRoute,"action") ){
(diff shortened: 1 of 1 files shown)


mpaluchowski:
Removing appending the default event name in ColdBox 3.8.1 has an adverse effect on the logic in `system.web.services.HandlerService`, line 258:

```
methodReceived 	= listLast(arguments.event,".");
```

It breaks when calling a handler inside a module, without stating an action explicitly. So, for instance, while the URL would be `/myModule:myHandler` and `defaultEventCheck` in `HandlerService` would correctly process the event to `myModule:myHandler.index`, right after the SES interceptor runs, the current event is `myModule:myHandler`, and in the above code `methodReceived` will be equal `myModule:myHandler` which is then set as the "method" in the EventHandlerBean instance.

Right afterwards this breaks in `HandlerService`, line 146:

```
if ( NOT oEventHandler._actionExists(arguments.ehBean.getMethod()) ){
```

which evaluates to `true`, since the said handler doesn't contain the method `myModule:myHandler`, and hence an invalid event exception is thrown.

PR #186, opened 2014-04-11
Update RequestService.cfc

mschneebeli:
Since runevent is not threadsafe because of the shared usage of the requestcontext. This approach offers each thread an individual scope where threads can run events without taking care of interference in the whole process.


lmajano:
Closing this as we discussed doing this as a request context decorator

PR #187, opened 2014-04-14
Update RequestContext.cfc

mschneebeli:
Another proposal for the thread issue. Much leaner compared to duplicating the whole context, but still have an own private collection for each thread....


lmajano:
closing as we discussed as a decorator

PR #200, opened 2014-09-22
Update RequestContext.cfc

mschneebeli:
Make the Private RequestContext ThreadSafe if desired. Just switch the setting: requestContextThreadDuplication to true and runEvent can call multiple handlers without interference of prc data.

It does only happen in threads anyway, so dont't be scared for the main thread's prc.


lmajano:
closing as we discussed as a decorator

PR #211, opened 2014-12-08
Update Logger.cfc

sigmaprojects:
Thread was unable to access thisAppender outside of it's own attributes scope (~~Reference~~* needs to be passed).

*My mistake, as Brad has pointed out it is not a reference but rather a deep copy.

### DIFF
--- a/system/logging/Logger.cfc
+++ b/system/logging/Logger.cfc
@@ -336,8 +336,8 @@ Description :
 								thisAppender.logMessage( logEvent );
 							} else {
 								// Thread this puppy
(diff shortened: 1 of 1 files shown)


lmajano:
I merged something manually to logbox, I have not seen any exceptions on it, can you check that out.

bdw429s:
Luis, can you elaborate on the "something" that was merged into LogBox?  Looking at the code, it seems like it wouldn't work for all the reasons already stated in this thread.  Perhaps the current approach works in one CFML engine, but not another?  I looked at this again because of a new thread on the ColdBox mailing list of people getting a similar error.

PR #212, opened 2015-02-02
Explicit Module view rendering

tgmweb:
I think I should be able to render a module view, and for it to use my conventions and not always be explicit. If I do renderView(view="myView",module="myModule") then it *always* takes the view from my external views location - this is never desired behavior for me - so I've added in here an option to override this for my use case and possibly others.

### DIFF
--- a/system/web/Renderer.cfc
+++ b/system/web/Renderer.cfc
@@ -146,7 +146,8 @@ component accessors="true" serializable="false" extends="coldbox.system.Framewor
 		numeric collectionStartRow="1",
 		numeric collectionMaxRows=0,
 		collectionDelim="",
-		boolean prePostExempt=false
+		boolean prePostExempt=false,
+		boolean renderExplicitModuleView=true
 	){
 		var viewCacheKey 		= "";
 		var viewCacheEntry 		= "";
@@ -165,7 +166,7 @@ component accessors="true" serializable="false" extends="coldbox.system.Framewor
 				arguments.module = event.getCurrentModule();
 			}
 		} else {
-			explicitModule = true;
+			explicitModule = arguments.renderExplicitModuleView;
 		}
 
 		// Rendering an explicit view or do we need to get the view from the context or explicit context?
--- a/system/web/services/HandlerService.cfc
+++ b/system/web/services/HandlerService.cfc
@@ -261,6 +261,28 @@ Description :
 		// Verify if this is a module call
 		if( find(":", arguments.event) ){
 			moduleReceived = listFirst(arguments.event,":");
+
+			// Add module handler over-rides
+      ParentHandlerReceived = "modules.#moduleReceived#.#HandlerReceived#";
+      customHandlerIndex = listFindNoCase(handlersList, ParentHandlerReceived);
+
+      // Check for conventions location
+      if ( customHandlerIndex ){
+        return HandlerBean
+          .setHandler(listgetAt(handlersList,customHandlerIndex))
+          .setMethod(MethodReceived);
+      }
+
+      // Check for external location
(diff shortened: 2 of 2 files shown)


tgmweb:
In addition - I've added my handler over-rides. This is a TOTAL BODGE - it works for my use case, but it would need to be re-written properly.

Essentially what it does is allows me to run custom module handlers that over-ride the external location. 

So in my custom location convention folder I have

mycustomsite
  handlers
    modules
      myModule
        customHandler.cfc

Hope that makes sense.

lmajano:
Why don't you use the "module" argument, that is an explicit rendering

tgmweb:
Yes I know that - see my code example in the first comment; that's exactly what I'm doing. My problem is, by specifying the module, it's **always** explicit (perhaps I'm using the wrong term here). 

```
renderView(view="myview", module="mymodule");
```
I'd expect the above to render the following view based on my conventions:
```
custom > views > modules > mymodule > myview.cfm 
```
Instead this is rendered:
```
modules > mymodule > views > myview.cfm 
```
Some people might want that behavior, and if so, passing an argument to renderView will maintain existing functionality whilst giving people (i.e me and possibly others) the option to not have it work like that.

tgmweb:
Test case here: http://dlri6gsip3fcx.cloudfront.net/tmp.zip

fire up commandbox and run this app. Then hit /main/ and /main/anotherTest.

You should get the same view, but you don't.

bdw429s:
Tom, can you help us understand your scenario? I ran your sample code but found it a little confusing because you set the views convention to "custom/views" and set the viewsExternalLocation to "views".  Is there a reason to flip them like that?  This makes your questions harder to understand since when you say "external location" you really mean the regular "views" folder.

Also, there is no such Coldbox convention to put a modules folder inside your views directory. Were you expecting Coldbox to look in /views/modules because you specified a module when you rendered the view?

tgmweb:
Hi Brad,

From what I understand, coldbox checks the "views" location first, then checks the external location. 

Therefore, my vanilla "off-the-shelf" app goes in the external location. Any client specific over-rides therefore go in the views location. I've always said this was confusing (the lookup order), but now I've kinda gotten used to it. So I don't think I've flipped anything. Perhaps the example is difficult to understand because there's only 1 view in each place. The majority of stuff goes in the external location.

If you want to override a module view - i.e a client wants a different module view than the vanilla app provides (lets say they don't want a remember me tickbox on the login page), you can only do that (as far as I can tell) using parentLookup and using the conventions I've used (putting the module inside the custom views folder). Coldbox finds the correct view virtually all the time. The only time is doesn't, is when using renderView() and adding the module argument. setView() always works.

Hopefully that explains it!

lmajano:
Ok, I am starting to get this. Basically, you want a way to say go render a view from a module but do the convention lookup as well.  As for the default behavior of setView() would be to do the lookup and module contexts.  This is is more for targeting a module but wanting conventions right?

If that's the case, I think we would need another argument in the renderView() method to do this, since the idea of explicit module rendering is for rendering exactly what you want.  How about:

```
renderView( view, module, boolean doLookups=false )
```

Which can default to false.  Thoughts?

tgmweb:
Luis - that's exactly it! Currently setView() always works, and if you do renderView() from within a module event/layout it **does** do the lookup - it's only when you do renderView() and specify the module it doesn't do the lookups.

I'm happy for an additional argument. The same goes for renderLayout() as well which I just noticed last night does the same thing.

Also if you go with an argument and do:
    renderLayout(layout="myLayout", module="myModule", view="myView", doLookups=true);

I'd say the doLookups argument should probably propagate through the the view too.

Thanks!

Tom.

lmajano:
I agree.  So if that's the case, then I will close this, so you can send it to the "development" branch and with those updates. Sorry for the long delay :)

PR #213, opened 2015-03-28
Add check for loadColdbox to bootstrap CB so ORM Entity unit tests work

HokieTT:
Also submitting changes for CommandBox so ORM Entity creation with tests will utilize this change.


lmajano:
Thanks, merged manually and ticket created: https://ortussolutions.atlassian.net/browse/COLDBOX-455

PR #215, opened 2015-06-26
COLDBOX-461

timmaybrown:
This seemed to fix the error I reported on JIRA:
https://ortussolutions.atlassian.net/browse/COLDBOX-461

Not sure if this the proper way to handle this or if it would make more sense to just use the accessor methods directly in the getDataPacket() method.

### DIFF
--- a/ApplicationTemplates/rest/models/Response.cfc
+++ b/ApplicationTemplates/rest/models/Response.cfc
@@ -30,6 +30,8 @@ component accessors="true" {
 		// Init complex properties
 		variables.messages 	= [];
 		variables.headers 	= [];
+		variables.error 	= THIS.getError();
+		variables.errorCode 	= javaCast('integer', THIS.getErrorCode());
 
 		return this;
 	}
--- a/ApplicationTemplates/rest/models/Response.cfc
+++ b/ApplicationTemplates/rest/models/Response.cfc
@@ -30,8 +30,8 @@ component accessors="true" {
 		// Init complex properties
 		variables.messages 	= [];
 		variables.headers 	= [];
-		variables.error 	= THIS.getError();
-		variables.errorCode 	= javaCast('integer', THIS.getErrorCode());
+		variables.error 	= getError();
+		variables.errorCode 	= javaCast('int', getErrorCode());
 
 		return this;
 	}


bdw429s:
is this a bug in CF11?  I'm unclear on whether CF isn't doing what we expected or if the sample app was actually wrong.

bdw429s:
Actually, looking at the code is this an issue with ColdFusion not obeying default values in properties.  I seem to remember Railo/Lucee using them and Adobe CF not.  Also, why the java cast?

timmaybrown:
When I run this same app via CF 11 and via commandbox embedded server the default values from the properties are not placed into the variables scope. The only properties that are in the variables scope on CF 11 are the `messages` and `headers` variables which are explicitly init'd in the init function.

I remember having to do this back on some of my Bean, DAO, gateway models running on CF, so this may be an error in CF but explicitly defining the variables in the init function allow the sample REST app to return the proper payload.

The javacast provides continuity to the response for the errorcode key which on Lucee is Quoted and on CF it returns as an un-quoted integer (see attached). Image on the right is the command box  2.x embedded server vs a local install of cf11 update 5.

luceevscf-serialization-wo-javacast

Without the variables init'd this is what the response on CF 11 started out as.
basehandlererrorwithoutinitederrorvars

lmajano:
I get it. This is a bug in ACF where the param defaults are not being respected.  Ok, sucks, more typing for us.

timmaybrown:
Looks like you already reported. Not sure of the implications of taking the approach in this blog post, but adding `persistent="true"` to the component declaration got rid of the error without the extra init'd variables.

http://cfsimplicity.com/23/using-cfproperty-to-create-default-cfc-property-values

lmajano:
Just added the inited variables. Persisten might cause issues with ORM. Thanks!

timmaybrown:
sounds good. ticket was opened also that can be closed as dupe of 462. https://ortussolutions.atlassian.net/browse/COLDBOX-461

lmajano:
Oops, did not see, it, sorry.




Luis Majano
CEO
Ortus Solutions, Corp
www.ortussolutions.comP/F: 1-888-557-8057


Direct: (909) 248-3408




ColdBox Platform: http://www.coldbox.org


ContentBox Platform: http://www.gocontentbox.org

Linked In: http://www.linkedin.com/pub/3/731/483







Social: twitter.com/ortussolutions | twitter.com/coldbox | twitter.com/lmajano | twitter.com/gocontentbox

PR #216, opened 2015-07-01
Fixed addAsset to better handle "tagged" js files

sistemanick:
addAsset should now recognize app.js?123 as an js file, not css.


lmajano:
Can you send this to the development branch please

PR #219, opened 2015-08-27
No default for ssl parameter, use only if provided

wphampton:
The *startForm* and *href* functions no longer supply a default value for
the *ssl* parameter.  When event.buildLink() is called the *ssl* parameter
is only passed if it was first supplied to the *startForm* or *href*
function.  Previously a value of false was default for the *ssl* parameter
and it forced *event.buildLink()* to build links as http:// instead of
letting it default the URL protocol as necessary.


wphampton:
This came out of a [DataBoss discussion](https://groups.google.com/a/ortussolutions.com/d/msg/databoss/hY0Z1l1IDLs/Fo-bUeWpBgAJ) with @bdw429s

lmajano:
Merged manually.

lmajano:
Gracias!!!

PR #221, opened 2015-09-01
Provider improvements

dswitzer:
I wanted to issue 2 separate pull requests, but I apparently mucked it up. I apologize for that.

I made two distinct changes here:

#### CFProvider

When a "cacheName" is specified, the configure() method now automatically registers the region in CF10+. Prior to CF9, you didn't need to create the region ahead of time to use it and I thought it made sense that CacheBox would automatically manage registering a region.

I also fixed an issue in getCachedObjectMetadata(), where there was no logic in place for when you had specified a cacheName. The current code just returns null instead of the metadata for the object.

#### MockProvider

I was writing some unit tests and needed the clearByKeySnippet() method of the MockProvider to actually work.

In order to accomplish this, I needed to change the getKeys() to return an array, but other than that, there really aren't any changes that should break anything.

### DIFF
--- a/system/cache/providers/CFProvider.cfc
+++ b/system/cache/providers/CFProvider.cfc
@@ -5,7 +5,7 @@ www.coldbox.org | www.luismajano.com | www.ortussolutions.com
 ********************************************************************************
 Author: Luis Majano
 Description:
-	
+
 This CacheBox provider communicates with the built in caches in
 the Adobe ColdFusion Engine.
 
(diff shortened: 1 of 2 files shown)


bdw429s:
Hi Dan, thanks for the pull.  Can you re-submit this to the development branch.  We don't accept pulls to master since that is the current stable release. Also, we need to work on the CF engine detection.  What you have now isn't Railo/Lucee friendly, and it breaks our rule of detecting specific engines/versions.  We abstract all such login into this class.  I'm linking to an older version of the file as most of it's logic has been removed in recent versions due to CF engine parity.

https://github.com/ColdBox/coldbox-platform/blob/3.8.1/system/core/util/CFMLEngine.cfc

So you would need to add an "isCertainFeature()" type of method and return true/false based on all the supported versions of Railo, Lucee, and Adobe CF.  I don't know if the CFMLEngine CFC is included in CacheBox either right off.  If not, the built would need modified to include it.

lmajano:
Everything that @bdw429s said :)

PR #222, opened 2015-09-10
LOGBOX-21 #comment Fixed bug & future proofed it #time 13m

ensemblebd:
See issue for details.
The canLog() function returns false 100% of the time due to text<->int comparison.
I fixed this and future proofed it.

### DIFF
--- a/system/logging/AbstractAppender.cfc
+++ b/system/logging/AbstractAppender.cfc
@@ -165,13 +165,14 @@ Description :
 	<cffunction name="canLog" output="false" access="public" returntype="any" hint="Checks wether a log can be made on this appender using a passed in level" colddoc:generic="Boolean">
 		<cfargument name="level" required="true" hint="The level to check if it can be logged in this Appender" colddoc:generic="text"/>
 		<cfscript>
-			loc={min=getLevelMin(),max=getLevelMax(),level=arguments.level};
-			keys=StructKeyList(loc);
-			for (i=1;i<=ListLen(keys);i++) { // force integers regardless of config value and parameter value, such that logic does not fail
-				key=ListGetAt(keys,i);
-				if (!isNumeric(loc[key])) loc[key]=this.logLevels.lookupAsInt(loc[key]);
+			LOCAL.values={min=getLevelMin(),max=getLevelMax(),level=arguments.level};
+			LOCAL.keys=StructKeyList(LOCAL.values);
+			LOCAL.i=1;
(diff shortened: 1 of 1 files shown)


bdw429s:
Thanks for the pull @ensemblebd !  Since the try catch would only mask errors by pretending they worked, can you simply remove that?  If a user passes in something crazy like a struct, we'll just let the function blow up so they know to change it.  No need to submit another pull, just an additional commit on your `logbox-21` branch will show up.

@lmajano Can you review as well?

system/logging/AbstractAppender.cfc
```diff
@@ -163,9 +163,17 @@ Description :
 
 	<!--- canLog --->
 	<cffunction name="canLog" output="false" access="public" returntype="any" hint="Checks wether a log can be made on this appender using a passed in level" colddoc:generic="Boolean">
-		<cfargument name="level" required="true" hint="The level to check if it can be logged in this Appender" colddoc:generic="numeric"/>
+		<cfargument name="level" required="true" hint="The level to check if it can be logged in this Appender" colddoc:generic="text"/>
 		<cfscript>
-			return (arguments.level GTE getLevelMin() AND arguments.level LTE getLevelMax() );
+			loc={min=getLevelMin(),max=getLevelMax(),level=arguments.level};
```
bdw429s: Variables are not locally scoped.  loc, i, and key at least.

ensemblebd:
Updated, let me know if any improvements can be made. Hopefully this helps a bit
Thanks again, love this product. Can't believe it's licensed free

lmajano:
Hi @ensemblebd thanks for the pull request.  However, the method always meant to take in *numeric* values and not text values. So basically, this pull request is *enhancing* the method to allow for *numeric* and *text* values.

ensemblebd:
Hello @lmajano thank you for your review and information. Though I would concur on that aspect, I also referenced a serious error in the Jira issue. If you supply it with a numerical value, and the configuration is using TEXTUAL values, it returns false 100% of the time, because you can't compare numbers with strings using "GTE" or "LTE". 

The original and current intent of this pull request is not to enhance anything, it is to fix a serious problem that prevented me from using LogBox. I would be happy to remove the enhancements if preferred, as I would much rather that, than to have the pull request categorized as a non-critical enhancement.

Thank you again, and appreciate your response and advise.

lmajano:
Thanks for the pull, merged manually! https://github.com/ColdBox/coldbox-platform/commit/b290a0d9931a40f22ae18a45ee1dcb984527c141

PR #230, opened 2015-10-11
CacheBox-33 Convert CacheBox's FIFO policy to script

animex:


bdw429s:
Thanks, please resubmit to develop!

PR #231, opened 2015-10-11
CACHEBOX-34 Convert CacheBox's LFU policy to script

animex:


bdw429s:
Thanks! @lmajano can you review and merge?

PR #232, opened 2015-10-16
CACHEBOX-33

bgthomas:


bdw429s:
Thx @bgthomas !  One small thing:
The `colddoc:generic="coldbox.system.cache.ICacheProvider` is a separate annotation for the `cacheProvider` argument.  It needs to get its own line like so:
```
@cacheProvider.colddoc:generic coldbox.system.cache.ICacheProvider
```
If you add this commit to your branch, it should automatically be part of the pull.

bgthomas:
Hi Brad,

I have added the line and committed it.

Let me know if there is anything else!

bdw429s:
Thanks @bgthomas!

PR #233, opened 2015-10-19
Fixes Cachebox-35

Tropicalista:
Fixes Cachebox-35

PR #234, opened 2015-10-19
Convert to script

amarras:


bdw429s:
Sorry @amarras it looks like we had more than one person working on this ticket but the ticket hadn't gotten assigned.  This pull is already complete via https://github.com/ColdBox/coldbox-platform/pull/233

Thanks for your contribution though.

PR #236, opened 2015-11-17
COLDBOX-485 fix for parent properties (recursive)

wpdebruin:
GetMetadata in getRelationshipMetaData function replaced by recursive
function collectAllProperties

### DIFF
--- a/system/core/dynamic/BeanPopulator.cfc
+++ b/system/core/dynamic/BeanPopulator.cfc
@@ -420,7 +420,9 @@ Description :
 		<cfscript>
 			var meta = {};
 			// get array of properties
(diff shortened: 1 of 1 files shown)


bdw429s:
Thanks for the pull.  ColdBox already has a utility class that will get inherited metadata for a CFC.  Is there a chance we can just re use this?
https://github.com/ColdBox/coldbox-platform/blob/master/system/core/util/Util.cfc#L210-L268

wpdebruin:
Probably yes. It didn't know it was there. It does recursion on ALL metadata, not only properties, but I think it will work. I will try and let you know

wpdebruin:
Ok, I reused the getInheritedMetadata method now, Just check if you like this one better :-)

PR #237, opened 2015-11-18
Fix parameters passed to AOP mixing exception

elpete:
The parameters passed to the AOP mixing exception were in the wrong
order.  Use named parameters to fix this issue in the future.

I uncovered this because the error was hidden under `type` in TestBox.  It also made it very hard to read when the error occurred in ColdBox. (The error message caused the debug box to span on horizontally for what seemed like forever.)

(See http://dev.elpete.com/2015/11/13/reserved-words-in-aop-aspects/ for more details)


lmajano:
Thanks merged manually

PR #239, opened 2015-11-27
slugify doesn't make nice German slugs

GunnarLieb:
German umlauts should be replace like this:
ä to ae
ü to ue
ö to oe
ß to ss

### DIFF
--- a/system/core/dynamic/HTMLHelper.cfc
+++ b/system/core/dynamic/HTMLHelper.cfc
@@ -394,7 +394,7 @@ Description :
 		<cfscript>
 			// Cleanup and slugify the string
 			var slug = lcase(trim(arguments.str));
-
+			slug = replaceList(slug,'#chr(228)#,#chr(252)#,#chr(246)#,#chr(223)#','ae,ue,oe,ss')
 			slug = reReplace(slug,"[^a-z0-9-\s#arguments.allow#]","","all");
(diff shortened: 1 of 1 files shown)


Tropicalista:
Good point, I have same problem with accented characters in italian: à,è,ì,ò,ù that are used in a lot of european language. This are removed by slugify. Probably a better solution would be to add two more arguments to the function with allowedList and replaceList.

Tropicalista:
We should also consider the approach of this SO ansers: http://stackoverflow.com/questions/9926507/coldfusion-convert-accented-regional-characters-to-plain-ascii and http://stackoverflow.com/questions/5806690/is-there-an-iconv-with-translit-equivalent-in-java/5807419#5807419

Tropicalista:
A simple gist: http://trycf.com/gist/25ef1499ea85058712f7

GunnarLieb:
The Gist you have posted doesn't do the right normalization for German. Germans would write AE for Ä and not A. I'm not sure about other languages

Tropicalista:
Yes, I don't know much about german language. The gist I posted is more related to the one that I know, Italian. I suppose that a lot of accented chars are also in Spanish, portuguese and french. I think could be a nice addition.

GunnarLieb:
Maybe we should split it up for each language? More lines of code but easier to maintain?

lmajano:
I don't mind adding them in for many use-cases .

PR #243, opened 2015-12-14
fix getValue() to throw exception correctly

henrylearn2rock:
without named arguments, exception will be wrong


bdw429s:
Thanks!

PR #244, opened 2016-01-08
Cachebox-37 Fix

eppinger:
use the logbox config from the cachebox config if in standalone moade


bdw429s:
Thanks for the pull @eppinger!  @lmajano this seems pretty straightforward.  Can you review it?

lmajano:
Thanks @eppinger Would be great to have some tests with this :)

lmajano:
@eppinger thanks, is there a way for some tests?

lmajano:
Thank you so much @eppinger I have added tests and merged manualy.

PR #245, opened 2016-01-21
Bean Populator - added additional argument to use the object's proper…

timmaybrown:
Bean Populator - added additional argument to use the object's properties to control key casing and only populate keys if they exist as properties. This allows the use of accessors=false and the use of onMissingMethod in combination with the trustedSetter argument to dynamically populate the bean object and control the casing of the object's properties which is especially useful to mitigate some of the key casing issues when building REST applications.


system/core/dynamic/BeanPopulator.cfc
```diff
@@ -442,4 +469,20 @@ Description :
 		<cfreturn createObject("component","coldbox.system.core.util.Util")/>
 	</cffunction>
 
+	<!--- Get the Target's properties as a struct with a key and value that are the same for easy access and setting of the key's casing --->
+	<cffunction name="getTargetPropertiesAsStruct" access="private" output="false" returntype="struct" hint="Return the property names for the target object as a struct with the key the same as the value">
```
lmajano: What about inheritance chains? Can you update this to search up the `extends` tree?

lmajano:
This is great. Is there a way to add some tests for this?

system/core/dynamic/BeanPopulator.cfc
timmaybrown: Happy to make any changes needed, just not entirely sure what that means, but if you have any direction as to what to do I'd love to try to figure it out.

timmaybrown:
Still trying to wrap my head around unit testing fully, but I'll see what I can figure out.

lmajano:
Thanks so much @timmaybrown , we like to challenge our pull requesters :)

lmajano:
Let us know @timmaybrown if you need help

system/core/dynamic/BeanPopulator.cfc
```diff
@@ -442,4 +469,22 @@ Description :
...
+		<cfargument name="target" required="true" type="any" />
+
+		<cfscript>
+			// get array of properties
+			var stopRecursions= [ "lucee.Component", "railo.Component", "WEB-INF.cftags.component" ];
```
bdw429s: FWIW, since these CFCs _are_ the base components, isn't this a little redundant?

timmaybrown: Likely. Just wasn't sure as I saw this line in use in the other instance where the `getInheritedMetaData()` method was used so I thought it might be needed. 

Just haven't had the time to work on some tests yet.

lmajano:
Let us know if you need help :)

bdw429s:
@lmajano Are we still waiting for tests cases or do we want to merge this as-is?  This pull has been here since January!

timmaybrown:
Sorry life happened a new little one... I really am trying to better understand writing tests and what needs to be tested for.. forgot I had submitted this. Any guidance you might have and I'd be happy to add the tests. Still haven't had time to fully wrap my head around all things bdd/testing.

bdw429s:
@timmaybrown I would recommend looking at the other unit tests and copying what they do.  The instructions for setting up the ColdBox unit tests are here:
https://github.com/ColdBox/coldbox-platform/blob/development/tests/readme.md
The ColdBox tests are here:
https://github.com/ColdBox/coldbox-platform/tree/development/tests/specs
But it looks like there's no tests yet for the bean populator.

timmaybrown:
I think I'm mostly figuring it out, but going to rework what I pushed since I just realized that the functions to "test" should/can be named to reflect on what they are testing and don't have to exactly match the function names in the actual CFC.  The orm-enabled version of the BeanPopulatorTest (which is just realized existed) clarified this fact and  a few other things (like using the expect() library vs $assert etc.)

I think I should just be making a new `Person.cfc` for my tests instead of mocking it. Mocking it adds a bunch of extra keys which for the tests I'm trying to write ( checking the serialized json is properly cased based on the property names )  makes the output of serializing the object contain a ton of extra keys. I think Lucee handles my scenario properly, but I added it since I was running into the issue in ACF 11.

I'll revisit / rework to add further coverage and continue to learn, but some things are finally making some more sense.

bdw429s:
@timmaybrown Looking good, and The travis CI build shows the tests are passing!  Let us know when you think you have the tests all ready to go and we'll get this thing merged.  The next thing you can help us with is getting the functionality added to the docs!
https://github.com/ColdBox/coldbox-docs

system/core/dynamic/BeanPopulator.cfc
lmajano: Yes, you can remove the stoprecursions, not needed

tests/specs/core/dynamic/BeanPopulatorTest.cfc
```diff
@@ -0,0 +1,66 @@
+<cfcomponent extends="coldbox.system.testing.BaseTestCase">
```
lmajano: This looks great, however, can you move some of these tests here: https://github.com/ColdBox/coldbox-platform/blob/development/tests/specs/orm-enabled/BeanPopulatorTest.cfc

The bean populator tests need to go into the `orm-enabled` specs folder in order for them to work.

lmajano: Yes, you got it

lmajano:
Hi @timmaybrown Can we pick up this pull request again. It lasted for too long. Can we revise this?

lmajano:
Hola @timmaybrown can we get an update to see if this is still relevant please?

PR #247, opened 2016-02-05
by default it should not mapp model CFCs, let user decide. Also reloa…

sanaullah:
…d performance improves (no need to check models folder if nothing there)

### DIFF
--- a/ApplicationTemplates/rest/handlers/BaseHandler.cfc
+++ b/ApplicationTemplates/rest/handlers/BaseHandler.cfc
@@ -74,7 +74,7 @@ component extends="coldbox.system.EventHandler"{
 		
 		// Global Response Headers
 		prc.response.addHeader( "x-response-time", prc.response.getResponseTime() )
-				.addHeader( "x-cached-response", prc.response.getCachedResponse() )
+				.addHeader( "x-cached-response", prc.response.getCachedResponse() );
 		
 		// Response Headers
 		for( var thisHeader in prc.response.getHeaders() ){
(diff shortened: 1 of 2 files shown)


bdw429s:
I don't think I agree with this. Firstly, this change would break most sites out there using modules with models in them since everyone is depending on the models to be auto mapped. One of the reasons we did this is because we want modules to be black box that contributes easily to the parent app (or other modules).  Looking at other languages like Node, we saw that there is no configuration necessary to install a module and immediately "require" the stuff inside of it in the rest of your app.  We want to push people to writing small, open source modules that provide utilities and logic that can be included easily into other apps and I think that auto mapping modules is the rule not the exception.

ApplicationTemplates/rest/handlers/BaseHandler.cfc
```diff
@@ -74,7 +74,7 @@ component extends="coldbox.system.EventHandler"{
 		
 		// Global Response Headers
 		prc.response.addHeader( "x-response-time", prc.response.getResponseTime() )
-				.addHeader( "x-cached-response", prc.response.getCachedResponse() )
+				.addHeader( "x-cached-response", prc.response.getCachedResponse() );
```
bdw429s: This is where your change went. You didn't create a branch for each pull request so all your commits to development ended up in this pull.  Since I'm closing this pull, you need to revert your changes in your repo, create a branch, add this semi-colon change back to there and resubmit a pull.  Git only allows you to have one pull request per branch.

bdw429s:
Closing this based on our conversations via E-mail.

PR #249, opened 2016-02-09
Add url option to execute for coldbox integration testing

elpete:
Allow a url to be passed to the `execute()` function.  The `BaseTestCase` will then parse the url and set up the SES interceptor.  This allows testing of routes as well as events.

### DIFF
--- a/system/interceptors/SES.cfc
+++ b/system/interceptors/SES.cfc
@@ -92,7 +92,7 @@ Description :
 			var key 		 = "";
 			var routedStruct = structnew();
 			var rc 			 = arguments.event.getCollection();
-			var cleanedPaths = getCleanedPaths( rc, arguments.event );
+            var cleanedPaths = getCleanedPaths( rc, arguments.event );
 			var HTTPMethod	 = arguments.event.getHTTPMethod();
 
 			// Check if disabled or in proxy mode, if it is, then exit out.
@@ -1006,7 +1006,7 @@ Description :
 	</cffunction>
 
 	<!--- Find a route --->
-	<cffunction name="findRoute" access="private" output="false" returntype="any" hint="Figures out which route matches this request and returns a routed structure">
+	<cffunction name="findRoute" access="public" output="false" returntype="any" hint="Figures out which route matches this request and returns a routed structure">
 		<!--- ************************************************************* --->
 		<cfargument name="action" 	 required="true"  hint="The action evaluated by the path_info">
 		<cfargument name="event" 	 required="true"  hint="The event object.">
--- a/system/testing/BaseTestCase.cfc
+++ b/system/testing/BaseTestCase.cfc
@@ -6,7 +6,7 @@
 * Base testing component to intergrate TestBox with ColdBox
 */
 component extends="testbox.system.compat.framework.TestCase" accessors="true"{
-	
+
 	/**
 	* The application mapping this test links to
 	*/
@@ -182,7 +182,7 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
 	* Builds an empty functioning request context mocked with methods via MockBox.  You can also optionally wipe all methods on it
 	* @clearMethods Clear methods on the object
 	* @decorator The class path to the decorator to build into the mock request context
-	* 
+	*
 	* @return coldbox.system.web.context.RequestContext
 	*/
 	function getMockRequestContext( boolean clearMethods=false, decorator ){
@@ -239,7 +239,7 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
 	}
 
 	/********************************************* APP RETRIEVAL METHODS *********************************************/
-	
+
 	/**
 	* Get the WireBox reference from the running application
 	* @return coldbox.system.ioc.Injector
@@ -259,7 +259,7 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
 	/**
 	* Get the CacheBox reference from the running application
 	* @cacheName The cache name to retrieve or returns the 'default' cache by default.
-	* 
+	*
 	* @return coldbox.system.cache.ICacheProvider
 	*/
 	function getCache( required cacheName="default" ){
@@ -276,7 +276,7 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
 
 	/**
 	* Get the RequestContext reference from the running application
-	* 
+	*
 	* @return coldbox.system.web.context.RequestContext
 	*/
 	function getRequestContext(){
@@ -285,7 +285,7 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
 
 	/**
 	* Get the RequestContext reference from the running application
-	* 
+	*
 	* @return coldbox.system.web.Flash.AbstractFlashScope
 	*/
 	function getFlashScope(){
@@ -297,7 +297,7 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
 	/**
 	* Setup an initial request capture.  I basically look at the FORM/URL scopes and create the request collection out of them.
 	* @event The event to setup the request context with, simulates the URL/FORM.event
-	* 
+	*
 	* @return BaseTestCase
 	*/
 	function setupRequest( required event ){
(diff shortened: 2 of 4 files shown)


elpete:
Fixes Jira ticket: https://ortussolutions.atlassian.net/browse/COLDBOX-492

system/testing/BaseTestCase.cfc
```diff
@@ -311,15 +311,17 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
...
 	* @eventArguments A collection of arguments to passthrough to the calling event handler method
 	* @renderResults If true, then it will try to do the normal rendering procedures and store the rendered content in the RC as cbox_rendered_content
-	* 
+	*
 	* @return coldbox.system.context.RequestContext
 	*/
 	function execute(
-		required event,
+		string event = "",
+        string url = "",
```
lmajano: Maybe this should be `URI` instead of `URL` as it technically it is a resource identifier not a link?

system/web/context/RequestContext.cfc
```diff
@@ -740,6 +740,15 @@ component serializable=false accessors="true"{
    		return this;
 	}
 
+    /**
+    * Get the rendered content from a ColdBox integration test
+    *
+    * @return cbox_rendered_content or an empty string
+    */
+    function getRenderedContent(){
```
lmajano: I think this needs to be in the `BaseTestCase`. That's where my instinct tells me it should go.  As it does not make sense for normal MVC operation.

elpete: I've noticed you use a way of mixing in functions. I'd love a crash course and then I can implement it here.

system/interceptors/SES.cfc
```diff
@@ -645,6 +647,17 @@ Description :
 		<cfreturn instance.enabled/>
 	</cffunction>
 
+    <!--- Getter/Setter PathsForTesting --->
+    <cffunction name="setIntegrationTestingPaths" access="public" output="false" returntype="any" hint="Set whether the interceptor is enabled or not and return itself">
```
lmajano: @elpete I know this will do the job, but my instinct tells me that we should not place any direct references to integration testing on production MVC code.  So now that you showed that this code works, maybe we can refactor it in such a way, where the `BaseTestCase` can manipulate the interceptor's path locations instead of adding a concrete integration point.

system/testing/BaseTestCase.cfc
```diff
@@ -333,13 +335,33 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
...
+            // if we were passed a url, parse it and prepare the SES interceptor for routing.
+            if ( arguments.url != "" ){
+                // separate the url into the url and the query string
+                var urlParts = explodeURL( arguments.url );
+
+                // add the query string parameters to the request context
+                getRequestContext().collectionAppend( urlParts.queryStringCollection );
+
+                // set the integration testing paths so SES routes will be recognized
+                getInterceptor( 'SES' ).setIntegrationTestingPaths(
```
lmajano: Instead of doing this, why don't we leverage MockBox to mock the `getCleanedPaths()` to return the `urlParts.url`?

```
prepareMock( getInterceptor( "SES" ) ).$( "getCleanedPaths", urlParts.url );
```

elpete: Great suggestion!  I've added that in the latest commit and it works beautifully.

system/interceptors/SES.cfc
elpete: This is removed thanks to your MockBox tip.

system/web/context/RequestContext.cfc
elpete: I implemented a rudimentary mixin here.  I'd love for you to take a look.

elpete:
@lmajano Thanks for the feedback!  I've updated the pull request with your suggestions.

system/testing/BaseTestCase.cfc
```diff
@@ -311,15 +311,17 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
...
 	* @eventArguments A collection of arguments to passthrough to the calling event handler method
 	* @renderResults If true, then it will try to do the normal rendering procedures and store the rendered content in the RC as cbox_rendered_content
-	* 
+	*
 	* @return coldbox.system.context.RequestContext
 	*/
 	function execute(
-		required event,
+		string event = "",
+        string uri = "",
```
bdw429s: Following the lead of the `setNextEvent()` method in ColdBox, would it be more appropriate to have a separate `queryString` argument?  Then you can provide it alongside an event or a route.

Is the `URI` arg meant to just be the route as opposed to the full URL with domain and everything?  The DocBox comment doesn't really specify.  I've never cared for `URI` as it's often used as a synonym for URL, and is at best confusing.  Any thoughts on calling it `route`  since that's basically what I think you're wanting to pass.

system/testing/BaseTestCase.cfc
```diff
@@ -485,4 +516,36 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
...
+        var uriParts = ListToArray( arguments.uri, '?' );
+
+        var queryParams = {};
+        if ( ArrayLen( uriParts ) > 1 ){
+            var queryParamsArray = ListToArray( uriParts[2], '&' );
+
+            for ( var queryStringPair in queryParamsArray ){
+                var pairParts = ListToArray( queryStringPair, '=' );
+                if ( ! StructKeyExists( queryParams, pairParts[1] ) ){
+                    queryParams[ pairParts[ 1 ] ] = pairParts[ 2 ];
```
bdw429s: Shouldn't this be using `URLDecode()`?

bdw429s: I think this will error if you have empty value such as `?foo=` since `pairParts[ 2 ]` won't exist.

system/testing/BaseTestCase.cfc
```diff
@@ -485,4 +516,36 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
...
+                }
+                else{
+                    if ( ! isArray( queryParams[ pairParts[ 1 ] ] ) ){
+                        var currentVal = queryParams[ pairParts[ 1 ] ];
+                        queryParams[ pairParts[ 1 ] ] = [];
+                        ArrayAppend( queryParams[ pairParts[ 1 ] ], currentVal );
+                    }
+
+                    ArrayAppend( queryParams[ pairParts[ 1 ] ], pairParts[ 2 ] );
```
bdw429s: Same bit about `URLDecode()` here and empty values.  But also, CFML's approach is to create a comma-delimited list, not an array.

system/testing/BaseTestCase.cfc
elpete: I could see a `queryString` argument making sense.  I could also see having both.

elpete: I'm not tied to the name `url` or `uri`.  The intent was to pass just the route portion (something like `/login` which fully qualified would be `http://localhost:8500/index.cfm/login`)

elpete:
I fixed the `explodeURI` method like you suggested @bdw429s, including passing a comma-separated list instead of an array when there are multiple of the same parameters.

elpete:
I added an optional query string parameter.  Both the `route` and `queryString` are being parsed now.

I do wonder if the `explodeRoute()` and `parseQueryString` functions belong in a Utility class.

lmajano:
@elpete This looks amazing now.

One observation though, Do you mind adding tests for them?

elpete:
Sure. Is there already a test for this component? I couldn't find it.

lmajano:
@elpete There is already an `integration` folder https://github.com/ColdBox/coldbox-platform/tree/development/tests/specs/integration

I would say create a new spec called `BaseTestIntegrations` and start off with a new story there.

elpete:
@lmajano Added the tests.  This should be good to go. :+1:

elpete:
Hey, @lmajano.  Anything else you need to finish this one up?

system/testing/BaseTestCase.cfc
```diff
@@ -485,4 +521,52 @@ component extends="testbox.system.compat.framework.TestCase" accessors="true"{
...
+    * Parses a query string into a struct
+    *
+    * @queryString a query string from a URI
+    *
+    * @return a struct of query string parameters
+    */
+    private struct function parseQueryString( required string queryString ){
+        var queryParams = {};
+
+        var queryParamsArray = ListToArray( URLDecode( arguments.queryString ), '&' );
```
bdw429s: The entire query string shouldn't be decoded at once.  This will break if the parameter name or value has an encoded `& `.  First, the key/value pairs should be broken up at the ampersands, then the keys and values should be separated at the `=` signs, and only then should both the key and value be decoded.

Consider the following query string:
```
type=book&name=Jack%20%26%20Jill
```

The line of code here will decode the ampersand in the `name` variable too soon, thus breaking the query string into 3 parts instead of 2.
```[
    "type=book",
    "name=Jack ",
    " Jill"
]```

PR #250, opened 2016-02-11
CF11 don't like direct copy struct to another variable, it loose refe…

sanaullah:
CF11 don't like directly assing struct to another variable, it loose reference as struct. so have to make a copy of it for assignment.

### DIFF
--- a/system/web/context/ExceptionBean.cfc
+++ b/system/web/context/ExceptionBean.cfc
@@ -35,7 +35,7 @@ component accessors="true"{
 		any extraMessage = "",
 		any extraInfo = ""
 	){
-		variables.exceptionStruct = duplicate(arguments.errorStruct);
+		variables.exceptionStruct = arguments.errorStruct;
 		if( !isStruct( variables.exceptionStruct ) ){
 			variables.exceptionStruct = {};
 		}
--- a/system/web/context/ExceptionBean.cfc
+++ b/system/web/context/ExceptionBean.cfc
@@ -35,7 +35,7 @@ component accessors="true"{
 		any extraMessage = "",
 		any extraInfo = ""
 	){
-		variables.exceptionStruct = arguments.errorStruct;
+		variables.exceptionStruct = duplicate(arguments.errorStruct);
 		if( !isStruct( variables.exceptionStruct ) ){
 			variables.exceptionStruct = {};
 		}
--- a/system/web/Renderer.cfc
+++ b/system/web/Renderer.cfc
@@ -305,16 +305,25 @@ component accessors="true" serializable="false" extends="coldbox.system.Framewor
 			variables._items = arguments.collectionMaxRows;
 		}
 
+		//local counter when using startrow is greater than one x values is reletive to lookup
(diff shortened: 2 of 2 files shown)


system/web/context/ExceptionBean.cfc
```diff
@@ -35,7 +35,7 @@ component accessors="true"{
 		any extraMessage = "",
 		any extraInfo = ""
 	){
-		variables.exceptionStruct = arguments.errorStruct;
+		variables.exceptionStruct = duplicate(arguments.errorStruct);
```
bdw429s: Can you explain specifically what this is doing.  This code actually used to duplicate the exception, but it causes issues because duplicate changes the underlying type of the object so we removed it.  Changing this back will likely cause regressions.  Here you can read one of the issues that duplicate caused people in the past:
https://groups.google.com/forum/#!topic/coldbox/yVZnyOyMOHU

sanaullah: Brad,

without duplicate in CF-11 Type,Message, Tag Context, StackTrace are empty

lmajano: What about in cf10?

sanaullah: CF-11 converting it into object 
--isObject(arguments.errorStruct) YES 
--isStruct(arguments.errorStruct) NO
Manually I can add info into struct 
```
if(isObject ) then
variables.exceptionStruct.Message = arguments.errorStruct.getMessage
--- etc
```
its really strange why CF-11 does it :( 

Lucee works fine 

I don't have CF-10 at the moment, just picking up code-base after a long time :)

lmajano:
I just remember what @bdw429s said as well.   When we had duplicate it introduced other issues. So not sure how we can tackle this. Thoughts?

sanaullah:
Luis,

What about below hack/fix for CF 10,11... if you think its ok then I will commit the code in github

```
ExceptionBean function init(
		any errorStruct = {},
		any extraMessage = "",
		any extraInfo = ""
	){
		variables.exceptionStruct = arguments.errorStruct;
		if( !isStruct( variables.exceptionStruct ) ){
			variables.exceptionStruct = {};
		}
		variables.extraMessage 	= arguments.extraMessage;
		variables.extraInfo 	= arguments.extraInfo;

		try {
			if(structIsEmpty(variables.exceptionStruct) && isObject(arguments.errorStruct)){
				var tErrorStruct = duplicate(arguments.errorStruct);
				variables.exceptionStruct["type"] = tErrorStruct.Type;
				variables.exceptionStruct["message"] = tErrorStruct.Message;
				variables.exceptionStruct["detail"] = tErrorStruct.Detail;
				variables.exceptionStruct["StackTrace"] = tErrorStruct.StackTrace;
				variables.exceptionStruct["TagContext"] = tErrorStruct.TagContext;
			}
		}
		catch(type e) {
			// suppress error (CF10,11 pass exception struct as object to this method )
		} 
		
		return this;
	}

```

PR #251, opened 2016-02-21
Global "/coldbox/system/modules" modules should be loaded before + COLDBOX-475

sanaullah:
Global "/coldbox/system/modules" modules should be loaded before other modules, otherwise dependencies fails...

COLDBOX-475 https://ortussolutions.atlassian.net/browse/COLDBOX-475
Rendering query-based view collections and collection will treated as struct in view files.

```
return renderView(view="home/people", collection=prc.qry, collectionAs='people', collectionStartRow=2, collectionMaxRows=5);

<cfoutput>
<h1>Title: #people.Title# (#_counter# of #_items#)</h1>
<p>Author: #people.FirstName#</p>
</cfoutput>
```
'''

### DIFF
--- a/system/web/Renderer.cfc
+++ b/system/web/Renderer.cfc
@@ -305,16 +305,25 @@ component accessors="true" serializable="false" extends="coldbox.system.Framewor
 			variables._items = arguments.collectionMaxRows;
 		}
 
+		//local counter when using startrow is greater than one x values is reletive to lookup
+		var _localCounter = 1;
 		for( x=arguments.collectionStartRow; x lte ( arguments.collectionStartRow+variables._items )-1; x++){
 			// setup local cvariables
-			variables._counter  = arguments.collection.currentRow;
-			variables[ arguments.collectionAs ] = arguments.collection;
+			variables._counter  = _localCounter;
+
+			var columnList = arguments.collection.ColumnList;
+			for(var j=1; j <= ListLen(columnList); j++){
+				variables[ arguments.collectionAs ][ListGetAt(columnList,j)] = arguments.collection[ListGetAt(columnList,j)][x];
+			}
+			
 			// prepend the delim
 			if ( variables._counter NEQ 1 ) {
 				buffer.append( arguments.collectionDelim );
 			}
+			
 			// render item composite
 			buffer.append( renderViewComposite( arguments.view, arguments.viewPath, arguments.viewHelperPath, arguments.args) );
+			_localCounter++;
 		}
 
 		return buffer.toString();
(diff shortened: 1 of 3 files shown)


bdw429s:
Sana, you need to be using local branches for each pull. All your committs on your dev branch are colliding.  If you look in the "changes" tab, you'll see too many commits are there.

Now, about this pull.  Shouldn't you be using `this.dependencies` in your module config to ensure that dependant modules get loaded first?

sanaullah:
Simple Answer: Big No :) 

We are pushing users to do lot more, secondly the global-modules concept is simple that these should be active first before any other module.

bdw429s:
Loading global modules first may work in your specific case, but what about a globally-installed module that depends on a "local" module provided in the app?  This.modules is the documented method to ensure dependant modules get loaded first, wherever they are.

sanaullah:
I am 100% sure now you have started the discussion to defeat " terminology and best practices ", based on your assumptions which will not take me anywhere.

I don't think you develop the apps that way as you describe in comments. :)

Also seems like you missed my point about global-modules path I am talking about "/coldbox/system/modules"

Well I am more than happy to cancel the PR :+1:

bdw429s:
To be clear, I don't see any problem with this pull request, but I don't think it solves what you think it does. It just reverses the problem scenario. There may be a valid argument that it stands to reduce the chance of modules loading in the incorrect order, but again, `this.dependencies` is the documented approach to ensure modules load in the correct order.  I gather that you don't like that answer, but I'll let you duke that out with @lmajano as he engineered that part of ColdBox :)

sanaullah:
Thanks Brad, 
Its similar situation of **autoMapModels** unless user decide not to, so if user decide to have dependency before global-module then he has to specify _this.dependencies_ otherwise system should follow the rules of loading modules.

lmajano:
Sana are you using the this.dependencies directive in module configuration?

In all reality it should not matter when modules get registered. The point that matters is when they are activated. 

That's why this.dependencies is used. Is that not working for you.

sanaullah:
Luis,

I know about `this.dependencies` but I am talking about "/coldbox/system/modules"  'Put all system wide modules in this folder.'  like i have all cbox-modules there and these should take the priority to get register first then other location's modules. 

At-least we are not misleading the user that global-modules are not loaded first. if user want to do something else then `this.dependencies` should be declared in module.

also there are apps which have more than 50 custom modules, so they have to specify `this.dependencies` global dependent module-name, I think its just a extra work or at-least we should stop calling it system-wide modules folder.

lmajano:
Ok. I would agree with that then. Since this does not alter activation j am ok with it. The only issue is that it brings other commits. Can you repull with another feature branch please? Also can you create a ticket for this please as an improvement. Thanks Sana.

PR #254, opened 2016-02-22
Coldbox 496

sanaullah:
[ Jira Ticket ](https://ortussolutions.atlassian.net/browse/COLDBOX-496)

Error detail, message empty on CF10 CF11


system/web/context/ExceptionBean.cfc
```diff
@@ -41,6 +41,20 @@ component accessors="true"{
 		}
 		variables.extraMessage 	= arguments.extraMessage;
 		variables.extraInfo 	= arguments.extraInfo;
+		
+		try {
+			if(structIsEmpty(variables.exceptionStruct) && isObject(arguments.errorStruct)){
```
lmajano: @sanaullah if we get here, we know it is this special case, why not just duplicate it instead:
`variables.exceptionStruct = duplicate( arguments.errorStruct )`

sanaullah: Hi Luis, 
it was originally with duplicate( arguments.errorStruct ) but changed later on. Brad had some issue so he can explain why he changed. At-least its broken/empty detail,message in error report page for CF10 + 11

lmajano: Yes definitely.  However, since we know we are in cf10+ and only ACF exhibits this behavior, let's use duplicate for it then.

lmajano:
Merged manually.

PR #255, opened 2016-02-25
Coldbox cfversion

sanaullah:
CF 10 , 11 changed method name to `getApplicationSettingsMap`


system/core/util/CFMappingHelper.cfc
```diff
@@ -14,7 +14,7 @@ component{
 
 		if ( listFirst( server.coldfusion.productVersion ) == 9 ) {
 			addMappingCF9( name, path);
-		} else if ( listFirst( server.coldfusion.productVersion ) == 10 ) {
```
lmajano: Sana, not sure why do this if the current method works for cf9,10,11 and 2016

PR #258, opened 2016-03-12
getFileInfo() requires "Absolute on-disk or in-memory path"

sanaullah:
`getFileInfo()` requires "Absolute on-disk or in-memory path". fixed for CF9+


lmajano:
I missed this, sorry

PR #264, opened 2016-04-02
Allow passing in of an object to use for the metadata inspection.

elpete:
This is related to our discussion on Slack.

I have a test component that is structured like so:

```cfc
component extends='Integrated.BaseSpecs.ColdBoxBaseSpec' appMapping='/quality' {

	function beforeAll() {
		super.beforeAll();
	}

	function afterAll(){
		// do your own stuff here
		super.afterAll();
	}
	
	function run() {
            // Run some tests....
        }
}
```

And the component it extends uses `BaseTestCase` like so:

```cfc
component extends='Integrated.BaseSpecs.AbstractBaseSpec' {

    property name="baseTestCase" type="coldbox.system.testing.BaseTestCase";

    function beforeAll() {
        super.beforeAll();

        variables.baseTestCase = new coldbox.system.testing.BaseTestCase();
        baseTestCase.beforeAll( this );
    }

    function afterAll() {
        super.afterAll();
    }

   // ...
}
```

I need to get the `appMapping` metadata property in to the `BaseTestCase`.  This pull requests allows an optional `metadataObj` to be passed in to the `beforeAll` and `setup` methods that would be used for metadata inspection instead of `this` (which is still default).

### DIFF
--- a/system/testing/BaseTestCase.cfc
+++ b/system/testing/BaseTestCase.cfc
@@ -40,8 +40,8 @@ component extends="testbox.system.compat.framework.TestCase"  accessors="true"{
 	* Inspect test case for annotations
 	* @return BaseTestCase
 	*/
-	function metadataInspection(){
-		var md = getMetadata( this );
+	function metadataInspection( metadataObj = this ){
+		var md = getMetadata( metadataObj );
 		// Inspect for appMapping annotation
 		if( structKeyExists( md, "appMapping" ) ){
 			variables.appMapping = md.appMapping;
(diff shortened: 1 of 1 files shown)


elpete:
I was *really* hoping to get this in the next version of ColdBox for a new testing module I'm working on and planning to unveil at Into the Box this year.  (It's also what my last pull request about running integrated tests using routes was for.)

lmajano:
Hmm

All the metadata is read into properties of the test case. They also have accessors. So you can get and set them just like any other property. Why don't you do that instead?

elpete:
Totally missed the `accessors=true`.  I'll use that.  Thanks.

PR #266, opened 2016-04-27
Elixir function for versioned assets

elpete:
New elixir function that returns the versioned path for a file.

This allows us to cache bust and version using ColdBox Elixir.
```cfm
<script src="#elixir( 'js/all.js' )#"></script>
// =>
<script src="/includes/build/js/all-dd0e0acd94.js"></script>
```

If we can't find the versioned path, we return the normal convention path.
```cfm
<script src="#elixir( 'css/all.css' )#"></script>
// =>
<script src="/includes/css/all.css"></script>
```

The `appMapping` is also taken into account if there is one:
```cfm
// appMapping = "/testapp"
<script src="#elixir( 'js/all.js' )#"></script>
// =>
<script src="/testapp/includes/build/js/all-dd0e0acd94.js"></script>
```

---

This pull request also introduces a new configurable `includesLocation` option.

`includesConvention` can be overwritten in the ColdBox.cfc config file in the conventions struct along with the other ColdBox conventions by specifying an `includesLocation` key.

The only place using the `includesConvention` so far is this new function, but it is available any where using the standard `controller.getSetting( "includesConvention", true )` call.

### DIFF
--- a/system/core/dynamic/HTMLHelper.cfc
+++ b/system/core/dynamic/HTMLHelper.cfc
@@ -1449,7 +1449,24 @@ Description :
 	</cffunction>
 
 	<!--- elixir --->
-	<cffunction name="elixir" output="false" access="public" returntype="string" hint="Finds the versioned path for an asset">
+	<cffunction name="elixir" output="false" access="public" returntype="void" hint="Adds the versioned path for an asset to the view">
+		<cfargument name="fileName" type="string" required="true" hint="The asset path to find relative to the includes convention directory"/>
+		<cfargument name="buildDirectory" type="string" required="false" default="build" hint="The build directory inside the includes convention directory"/>
+		<cfargument name="sendToHeader" type="boolean" required="false" default="true" hint="Send to the header via htmlhead by default, else it returns the content"/>
+		<cfargument name="async" type="boolean" required="false" default="false" hint="HTML5 JavaScript argument: Specifies that the script is executed asynchronously (only for external scripts)"/>
+		<cfargument name="defer" type="boolean" required="false" default="false" hint="HTML5 JavaScript argument: Specifies that the script is executed when the page has finished parsing (only for external scripts)"/>
+		<cfscript>
+			addAsset(
+				elixirPath( arguments.fileName, arguments.buildDirectory ),
+				arguments.sendToHeader,
+				arguments.async,
+				arguments.defer
+			);
+		</cfscript>
+	</cffunction>
+
+	<!--- elixirPath --->
+	<cffunction name="elixirPath" output="false" access="public" returntype="string" hint="Finds the versioned path for an asset">
 		<cfargument name="fileName" type="string" required="true" hint="The asset path to find relative to the includes convention directory"/>
 		<cfargument name="buildDirectory" type="string" required="false" default="build" hint="The build directory inside the includes convention directory"/>
 		<cfscript>
--- a/system/core/dynamic/HTMLHelper.cfc
+++ b/system/core/dynamic/HTMLHelper.cfc
@@ -1454,10 +1454,10 @@ Description :
 		<cfargument name="buildDirectory" type="string" required="false" default="build" hint="The build directory inside the includes convention directory"/>
 		<cfscript>
 			var includesConvention = controller.getSetting( "includesConvention", true );
-			var appMapping = controller.getSetting( "appMapping" );
-			var filePath = expandPath("#appMapping#/#includesConvention#/#arguments.buildDirectory#/rev-manifest.json");
+			var mapping = event.getCurrentModule() != "" ? event.getModuleRoot() : controller.getSetting( "appMapping" );
+			var filePath = expandPath("#mapping#/#includesConvention#/#arguments.buildDirectory#/rev-manifest.json");
 
-			var href = "#appMapping#/#includesConvention#/#arguments.fileName#";
+			var href = "#mapping#/#includesConvention#/#arguments.fileName#";
 			
 			if ( ! fileExists( filePath ) ) {
 				return href;
@@ -1473,7 +1473,7 @@ Description :
 				return href;
 			}
 
-			return "#appMapping#/#includesConvention#/#arguments.buildDirectory#/#json[ arguments.fileName ]#";
+			return "#mapping#/#includesConvention#/#arguments.buildDirectory#/#json[ arguments.fileName ]#";
 		</cfscript>
 	</cffunction>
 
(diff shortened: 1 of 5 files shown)


lmajano:
@elpete I will add comments inline. However, I think the best place for the function is in the `HTMLHelper` class and not in the renderer.

`html.elixir()`

This way we can add more convention things in the html helper which does already `addAsset()` and others.

system/web/Renderer.cfc
```diff
@@ -694,6 +694,29 @@ component accessors="true" serializable="false" extends="coldbox.system.Framewor
 		return locateView(arguments.view);
 	}
 
+	public string function elixir( required string fileName, string buildDirectory = 'build' ) {
```
lmajano: 1. Please move to the HTML Helper
2. Add documentation for function and arguments

system/web/Renderer.cfc
```diff
@@ -694,6 +694,29 @@ component accessors="true" serializable="false" extends="coldbox.system.Framewor
...
+		var includesConvention = controller.getSetting( "includesConvention", true );
```
lmajano: You will need to also make this smarter for modules.  So basically, you need to check if you are in a module call or not.

You can use the `event.getCurrentModule()` and if that is empty, that means you go to the parent app, else you can go into the module using: `event.getModuleRoot()` + the convention `includes`.

lmajano:
You will also most likely need to add the `includesConvention` to the Module service, so modules can use it or override it.

elpete:
Should `elixir` return the path or add the asset?

I see three ways this could go:
1) `elixir` returns a path.  It takes an optional argument to add it as an asset.
```cfm
<cfoutput>
#html.elixir( path = 'css/all.css', addAsset = true )#
</cfoutput>
```
2) `elixir` returns a path.  You can add it as an asset like so:
```cfm
<cfoutput>
#html.addAsset( html.elixir( 'css/all.css' ) )#
</cfoutput>
```
3) `addAsset` can take an optional argument to look up the versioned path via `elixir`. <--- Not a fan of this one.
```cfm
<cfoutput>
#html.addAsset( asset = 'css/all.css', versioned = true )#
</cfoutput>
```
4) We split `elixir` into `elixir` (which adds an asset) and `elixirPath` (which returns the path).
```cfm
<cfoutput>
#html.elixir( 'css/all.css' )#
<!--- same as... --->
#html.addAsset( html.elixirPath( 'css/all.css' ) )#
</cfoutput>
```

Of those, I prefer **2** or **4**.  Thoughts?

system/web/Renderer.cfc
elpete: So...something like this?
```cfc
var mapping = event.getCurrentModule() != "" ? event.getModuleRoot() : controller.getSetting( "appMapping" );

var href = "#mapping#/#includesConvention#/#arguments.path#";
```
Or does the `appMapping` need to be there regardless, more like:
```cfc
var appMapping = controller.getSetting( "appMapping" );
var moduleMapping = event.getCurrentModule() != "" ? event.getModuleRoot() & "/" : "";

var href = "#appMapping#/#moduleMapping##includesConvention#/#arguments.path#";
```

system/core/dynamic/HTMLHelper.cfc
```diff
@@ -1448,6 +1448,35 @@ Description :
 		</cfscript>
 	</cffunction>
 
+	<!--- elixir --->
+	<cffunction name="elixir" output="false" access="public" returntype="string" hint="Finds the versioned path for an asset">
+		<cfargument name="fileName" type="string" required="true" hint="The asset path to find relative to the includes convention directory"/>
+		<cfargument name="buildDirectory" type="string" required="false" default="build" hint="The build directory inside the includes convention directory"/>
+		<cfscript>
+			var includesConvention = controller.getSetting( "includesConvention", true );
+			var appMapping = controller.getSetting( "appMapping" );
```
elpete: So...something like this?
```cfc
var mapping = event.getCurrentModule() != "" ?
    event.getModuleRoot() :
    controller.getSetting( "appMapping" );

var href = "#mapping#/#includesConvention#/#arguments.path#";
```
Or does the `appMapping` need to be there regardless, more like:
```cfc
var appMapping = controller.getSetting( "appMapping" );
var moduleMapping = event.getCurrentModule() != "" ?
    event.getModuleRoot() & "/" :
    "";

var href = "#appMapping#/#moduleMapping##includesConvention#/#arguments.path#";
```

lmajano:
I like option #4.

One method `elixir` for adding the asset.  and another for the path `elixirPath`

system/core/dynamic/HTMLHelper.cfc
lmajano: No. The `moduleMapping` has everything you need. SO the first part looks good.

system/web/services/ModuleService.cfc
```diff
@@ -248,7 +248,8 @@ I oversee and manage ColdBox modules
 						handlersLocation 	= "handlers",
 						layoutsLocation 	= "layouts",
 						viewsLocation 		= "views",
-						modelsLocation      = "models"
+						modelsLocation      = "models",
+						includesLocation    = "includes"
```
elpete: Is there any where else in modules that needs the `includesLocation`?  I couldn't see anywhere it was used....

lmajano: THis is is. Single place :)

elpete:
@lmajano I believe this pull request is complete, then!

system/core/dynamic/HTMLHelper.cfc
```diff
@@ -1448,6 +1448,52 @@ Description :
...
+			);
+		</cfscript>
+	</cffunction>
+
+	<!--- elixirPath --->
+	<cffunction name="elixirPath" output="false" access="public" returntype="string" hint="Finds the versioned path for an asset">
+		<cfargument name="fileName" type="string" required="true" hint="The asset path to find relative to the includes convention directory"/>
+		<cfargument name="buildDirectory" type="string" required="false" default="build" hint="The build directory inside the includes convention directory"/>
+		<cfscript>
+			var includesConvention = controller.getSetting( "includesConvention", true );
```
lmajano: This should be `includesLocation` not convention as per the setting.

lmajano:
One minor change needed.

elpete:
Ready to go, @lmajano.

lmajano:
Thanks @elpete fantastic work.  Quick question. Where should we document these functions?

PR #271, opened 2016-06-25
Add leading slash when including a module's routes files

elpete:
I plan to add tests for this, but I wanted to put this here so I didn't forget.

Without this leading slash, I get the following error:

```
Error importing routes configuration file: Page /coldbox/system/interceptors/modules_app/api_v1/config/routes.cfm [/Users/eric.peterson/Code/github/elpete/my_app/coldbox/system/interceptors/modules_app/api_v1/config/routes.cfm] not found
```

It's trying to pull the routes from the local interceptors folder instead of the root of the app.  Adding the leading slash fixes this.


lmajano:
Do i wait?

elpete:
Yes. I'll add a test case on Monday. 👍




On Sat, Jun 25, 2016 at 7:01 PM -0600, "Luis Majano" <notifications@github.com> wrote:












Do i wait?

elpete:
Sorry about all the commits.  I can't get the server to run on my local right now. 😥

elpete:
Okay, it's looking like a legitimate error somewhere when adding the routes like this.  I'll have to wait until I can get it running locally again.  Glad we waited for the tests!

PR #273, opened 2016-07-07
Use the default event if the route passed in is the base route (`/`)

elpete:
This lets you test like so:

```cfc
var event = execute( route = "/", renderResults = true );
```

Which, by default, would execute `Main.index`.

### DIFF
--- a/system/testing/BaseTestCase.cfc
+++ b/system/testing/BaseTestCase.cfc
@@ -340,6 +340,12 @@ component extends="testbox.system.compat.framework.TestCase"  accessors="true"{
         }
 
         try{
+        	// If the route is for the home page, use the default event in the config/ColdBox.cfc
+        	if ( arguments.route == "/" ){
+        		arguments.event = getController().getSetting( "defaultEvent" );
+        		arguments.route = "";
+        	}
+
             // if we were passed a route, parse it and prepare the SES interceptor for routing.
             if ( arguments.route != "" ){
             	// enable the SES interceptor


bdw429s:
Sweet.  Does this fire all the "normal" bits of the SES interceptor?  Someone recently ran into an issue where they (understandably) expected the full ColdBox routing mechanism to kick in during integration testing.

elpete:
It depends on what they mean by full.  It doesn't follow redirects, for one.

All this did was see if the route was "/" and if it was executed the defaultEvent.

elpete:
I'm curious about what didn't work as expected for this person.  Do you have more details?

bdw429s:
It was Andrew Dixon at dev.O.  He had set up his `routes.cfm` file to have a different method for each HTTP verb.  Basically, POST ran his regular handler action and all others ran an InvalidEvent action.  He expected to be able to test that in his integration tests by hitting it with the wrong HTTP verb and returning his invalidEvent view.  He wasn't running his tests via the route though-- just hitting the event directly, which of course, bypasses everything in the routes file as well as all logic in the SES interceptor.  I remembered seeing a pull from you a while ago to help with this but I couldn't remember if it got merged and what version of ColdBox/TestBox it was part of.  Seeing this pull today reminded me of it.  If it's something that's live, are there docs for it anywhere?

elpete:
The changes to the `execute` method including the `route` parameter are [documented here.](http://coldbox.ortusbooks.com/content/testing/integration_testing/execute.html)

You are correct that only routes go through the SES interceptor.  The goal is definitely to simulate the full ColdBox routing mechanism with `route`, so if it doesn't a ticket should be filed.

lmajano:
Excellent. Can you create the ticket for this so I can link it please

elpete:
Completes https://ortussolutions.atlassian.net/browse/COLDBOX-531

PR #274, opened 2016-07-07
Bug with module routes file

elpete:
Fix where ColdBox would look for a module's routes file from the current directory instead of the module root.


lmajano:
Great job. Is there a ticket for this to link?

elpete:
So, some quick discussion on this one, this had to do with `modules_app` which I was having issues with as well. (See https://github.com/ColdBox/coldbox-platform/pull/275)

I don't think this was a bug nor is this desired behavior (now that I understand what the external locations are).  With that in mind, I'm closing this PR.

PR #275, opened 2016-07-07
Add app mapping to modules external location

elpete:
I asked this question on Slack, so this might be an unneeded pull request, but here goes.

I was recently trying to use a module in my integration tests.  This usually works when extending `coldbox.system.testing.BaseTestCase`.  In this case, it wasn't.  After some debugging, I found that the module wasn't being loaded because it was in my `modules_app` folder which was loaded from the `ModulesExternalLocation` configuration option and did not have the test app mapping of `/root`.

If this is the desired behavior, this pull request isn't necessary, but maybe a documentation one is.  I have been using the `modules_app` folder as a sort of private modules directory.  ForgeBox and other public modules go in the `modules` directory and are ignored from source control and my private app modules go in the `modules_app` directory and are committed to source control.  I have even heard that approach talked about from Ortus people.

So, if `ModulesExternalLocation` is not supposed to have the `appMapping` appended to it, what is the recommended approach to loading all these modules up in a integration test environment?

### DIFF
--- a/system/interceptors/SES.cfc
+++ b/system/interceptors/SES.cfc
@@ -248,7 +248,14 @@ Description :
 					// prepare module pivot
 					instance.withModule = arguments.module;
 					// Include it via conventions using declared route
-					includeRoutes(location=mConfig[arguments.module].mapping & "/" & mConfig[arguments.module].routes[x]);
+
+					// add a slash to the beginning of the string if it doesn't already have one
+					var moduleMapping = mConfig[arguments.module].mapping;
+					if ( left( moduleMapping, 1 ) != "/" ) {
+						moduleMapping = "/" & moduleMapping;
+					}
+
+					includeRoutes(location=moduleMapping & "/" & mConfig[arguments.module].routes[x]);
 					// Remove pivot
 					instance.withModule = "";
 				}
--- a/system/testing/BaseTestCase.cfc
+++ b/system/testing/BaseTestCase.cfc
@@ -340,6 +340,12 @@ component extends="testbox.system.compat.framework.TestCase"  accessors="true"{
         }
 
         try{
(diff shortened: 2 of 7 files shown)


bdw429s:
If I understand this, your change would require module external locations to always be inside the web root.  That is not desired.  It's up to the devs to ensure that the external module locations is a CFC paths that will resolve correctly, either in their main app or in their unit tests.  An external module location could point anywhere on your hard drive, so appending an app mapping a module path outside the web root would break it.  This is why we usually create some form of a `/root` mapping that points to the web root, so CFC paths that you want to be homed to the app root will always resolve, but that's not a requirement or expectation for external module locations..

elpete:
Okay.  I suspected that it was really just me not understanding what the "external locations" were.

bdw429s:
For what it's worth, now that we're pushing for people to use `/modules_app` as the convention for modules not installed from CommandBox, I'm tempted to say it should be a built-in convention for Coldbox.  @lmajano?  This wouldn't affect how other external locations work, but would have helped in this scenario.

elpete:
I think a separate convention would be a good idea. As you can see, I was very confused at first. The concept makes sense now.

lmajano:
I think maybe defaulting the external locations with it is a good step. Thoughts?

Luis Majano
CEO
Ortus Solutions, Corp
www.ortussolutions.com
P/F: 1-888-557-8057




On Thu, Jul 7, 2016 at 7:50 PM -0700, "Eric Peterson" <notifications@github.com> wrote:










I think a separate convention would be a good idea. As you can see, I was very confused at first. The concept makes sense now.

elpete:
Defaulting with what exactly? `/root/modules_app` and a `/root` mapping in `/Application.cfc`?

PR #276, opened 2016-07-12
Make getAppendersMap public

elpete:
Allows easier programmatic configuration of LogBox (i.e. for use with modules).


elpete:
Can someone re-run this build?  It looks like a network error.

bdw429s:
I don't see a way to re-run the build.  I also don't know if I have any permissions to do so on Travis.  @jclausen  @lmajano Is there a way I can re-run the build?  Do I need set up in Travis?

PR #277, opened 2016-07-16
Fixing response URL

andrew-dixon:
In some situations the response from the `getHTMLBaseURL` is returns with a double slash at the end, e.g.:

`http://www.mydomain.com//`

instead of:

`http://www.mydomain.com/`

This change fixes that by changing the replace to a regex replace and adding an optional slash after `index.cfm`.

### DIFF
--- a/system/web/context/RequestContext.cfc
+++ b/system/web/context/RequestContext.cfc
@@ -770,7 +770,7 @@ component serializable=false accessors="true"{
 	* Get the HTML base URL that is used for the HTML <base> tag. This also accounts for SSL or not.
 	*/
 	string function getHTMLBaseURL(){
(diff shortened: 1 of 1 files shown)


andrew-dixon:
@lmajano I didn't, but will take a look later...

andrew-dixon:
@lmajano Did you fix the test? Sorry hadn't had time to look yet.

lmajano:
It was ok all the time.  It was a timeout issue

Luis F. Majano
CEO
Ortus Solutions, Corp
www.ortussolutions.com

ColdBox Platform: http://www.coldbox.org
Linked In: http://www.linkedin.com/pub/3/731/483
Blog: http://www.luismajano.com
IECFUG Manager: http://www.iecfug.com
Social: twitter.com/lmajano facebook.com/lmajano

PR #278, opened 2016-07-18
Coldbox 532

bdw429s:
`event.getHTTPContent()` doesn't work on binary encoding


bdw429s:
https://ortussolutions.atlassian.net/browse/COLDBOX-532

PR #280, opened 2016-08-10
Fixed compatibility for CF10 , CF11

sanaullah:


lmajano:
thanks sana, I just prepended it.  Made no sense to check.  The module service takes care of it.

PR #281, opened 2016-08-10
Coldbox 536

bdw429s:
Remove whitespace coming from the interceptor service by converting tags to script


bdw429s:
https://ortussolutions.atlassian.net/browse/COLDBOX-536

lmajano:
How would this work if the function states it has no output?

bdw429s:
@lmajano I don't understand your question. This is the exact same function, just in script.  Look in the function comment. Output is set to true.

lmajano:
Had to close it as I could not merge it.  Also, I did see now the output. I was missing that.

bdw429s:
Doing a manual merge, I assume?

bdw429s:
@lmajano Are you going to add these changes??

lmajano:
done

PR #282, opened 2016-08-19
Full null support

elpete:
Running all the tests never completes with full null support turned on.

I don't have the tools or know-how necessary to profile / debug that.

Everything passes with full null support off, though. 👍

### DIFF
--- a/system/Bootstrap.cfc
+++ b/system/Bootstrap.cfc
@@ -211,7 +211,7 @@ component serializable="false" accessors="true"{
 						renderedContent = cbController.getDataMarshaller().marshallData( argumentCollection=renderData );
 					}
 					// Check for Event Handler return results
-					else if( structKeyExists( refResults, "results" ) ){
+					else if( structKeyExists( refResults, "results" ) && ! isNull( refResults.results ) ){
 						renderedContent = refResults.results;
 					}
 					// Render Layout/View pair via set variable to eliminate whitespace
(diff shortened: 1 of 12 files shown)


elpete:
The tests that failed run fine locally. ¯\_(ツ)_/¯

bdw429s:
That error looks like something @lmajano and I were just looking at the other day.  I think Luis fixed it in this commit:
https://github.com/ColdBox/coldbox-platform/commit/116f1f04986b6253d26a9d6cd88ef5c95e9639c8
Have you re-based your branch from `development` recently?

elpete:
Just today.  This branch is completely up to date.

elpete:
Can someone re-run this build?  It looks like it failed for a random reason (`can't create directory /logs`).

bdw429s:
The tests are finally passing on this and at least one person has tested it and says it works for them.  I'm merging this now and we'll keep an eye out for any problems it may cause.

lmajano:
I am going through it and have some observations.  This code is repeated with the same pattern, checking the struct for existence and then a null check,

```
if( structKeyExists(refLocal,"target") && ! isNull(refLocal.target) ){
```

Why couldn't we just check for nullness:

```
if( ! isNull( refLocal.target ) ){
```

Intead of doing two checks on everything.

lmajano:
Also, the version was not bumped, so it re-build itself in staging as 4.3

elpete:
I see no reason why we couldn't just use the check for nullness.  Seems to work in my testing.

lmajano:
Do you mind Eric updating another pull? I am going through the other requests today

PR #283, opened 2016-08-26
dot notation fix

joeltobey:
This helps modules with multi-folder mapping (i.e. foo/bar). The Binder will now replace the folder / with a dot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 /settings.xml
 .project
+.settings/
 test-harness/logs/*.log
 test-harness/logs
 test-harness/test/logs
--- a/system/ioc/config/Binder.cfc
+++ b/system/ioc/config/Binder.cfc
@@ -293,7 +293,7 @@ Description :
  			</cfif>
 
 			<!--- Remove .cfc and /\ with . notation--->
-			<cfset thisTargetPath = arguments.packagePath & "." & reReplace( replaceNoCase( qObjects.name, ".cfc", ""), "(/|\\)", ".", "all")>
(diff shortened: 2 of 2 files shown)


bdw429s:
@joeltobey Can you include a quick example of what code doesn't work without this fix and now does?  Also, is there a ticket for this to cross reference?

joeltobey:
@bdw429s When my module model was mapped, the return type didn't match what I had for my init(). For instance, I had a service at /modules_app/foo-bar/models/MyService.cfc. In the ModuleConfig, I made my cfmapping = "for/bar". My service's init() method's signature returned "foo.bar.models.MyService". However, it was failing because the component was actually of type "foo/bar.models.MyService". I noticed the binder was replacing the slashes with dots (e.g. so that models/MyService would turn into models.MyService), but it wasn't taking into account if the packagePath had slashes in them as well. My fix addresses this use case.

I didn't see a ticket on this. I hope that's okay.

bdw429s:
Ahh, I see.  So, the CFCs are getting created fine, it just confuses CF's internal type checking.  I knew that CF will allow slashes or spaces internally when creating CFCs so that's why  I was confused.  This sounds like a CF bug really-- their type checking has always kind of sucked to the point where I usually don't even use it.

Regardless, I don't see any issue with this fix as it helps make things more consistent inside WireBox. @lmajano do you have any feedback?  Note, the unit tests in travis failed, but it's an unrelated error we fixed recently:
> Error: Cannot find a jsessionid, URLToken or cfid/cftoken in the cookie scope. Please verify
I assume @joeltobey 's branch isn't rebased?

lmajano:
merged manually, thanks

PR #284, opened 2016-08-26
Test runevent

sanaullah:
example code to reproduce the runevent issues (Version CF-11)

locally I reproduce it in test-harness 
http://cb.local/test-harness/index.cfm/testevent


lmajano:
fixed

PR #285, opened 2016-09-13
Bubble up an invalid HTTP method

evagoras:
Instead of throwing a 405 in the SES itself, why not let it bubble up to the function `onInvalidHTTPMethod` in a base Controller, which would allow you to specify a custom controller function to handle this in anyway you can?

For example in our REST API, we have that function in our base Controller:
```
function onInvalidHTTPMethod
( event, rc, prc, faultAction, eventArguments )
{
	// Setup Response
	prc.response = wirebox.getInstance( "beans.response@v1" )
		.setError( true )
		.setErrorCode( 405 )
		.setStatusCode( 405 )
		.setStatusText( "Method Not Allowed" )
		.setData( { "Message" = "The requested resource does not support the HTTP method '#event.getHTTPMethod()#'." } );
	// Render Error Out
	event.renderData
	(
		type		= prc.response.getFormat(),
		data 		= prc.response.getData(),
		contentType = prc.response.getContentType(),
		statusCode 	= prc.response.getStatusCode(),
		statusText 	= prc.response.getStatusText(),
		location 	= prc.response.getLocation(),
		isBinary 	= prc.response.getBinary()
	);
```

### DIFF
--- a/system/interceptors/SES.cfc
+++ b/system/interceptors/SES.cfc
@@ -139,10 +139,8 @@ Description :
 						}
 					}
 					else{
-						getUtil().throwInvalidHTTP(className="SES",
-												   detail="The HTTP method used: #HTTPMethod# is not valid for the current executing resource. Valid methods are: #aRoute.action.toString()#",
-										 		   statusText="Invalid HTTP method: #HTTPMethod#",
-										 		   statusCode="405");
+						// Bubble up the invalid to the controller
+						aRoute.action = "onInvalidHTTPMethod";
 					}
 				}
 				// Create routed event


lmajano:
The only problem with this approach is that if the action does not exist, then it will thrown an exception that the `onInvalidHTTPmethod` function does not exist.

lmajano:
Thought: What if we add a `isInvalidHTTPMethod()` in the request context that could tell the eco system that the incoming request has an invalid http method.  Then we could do listeners as well like: `onInvalidHTTPMethod` interceptor

evagoras:
That could work. Also, another thought, what if there was an option in the ColdBox.cfc settings, something like coldbox.settings.onInvalidHTTPMethod, which you could set as a pointer to a controller action? Then the SES logic could check first if that is defined and run it, otherwise do what it does now (throw a 405).

lmajano:
Ok, how about this. As I think we need different levels.

- Do your action set to `onInvalidHTTPMethod` if not found
- Update Controller, if the action does not exist (onInvalidHTTPMethod) then look for the `invalidHTTPMethodHandler` setting. If no setting, then throw exception

lmajano:
Ok, merged this manually, and added full cascading support. THanks sso much

evagoras:
Awesome! That looks great.

PR #287, opened 2016-09-20
Update version number to latest in README

elpete:


bdw429s:
@lmajano It looks like the readmes for WireBox and CacheBox don't even have a version in there.  Should we just remove it from Logbox to be consistent?  The `box.json` for each package is really the place for the version number now.

PR #289, opened 2016-10-05
Parse parent settings by convention for modules

elpete:
By convention, look for a struct in the `config/ColdBox.cfc` parent configuration's `moduleSettings` struct matching the module's namespace.
If it is found and the `this.parseParentSettings` flag is `true` (default), then merge the parent config struct in to the module settings struct.
Then set the new module settings struct as the settings in the parent configuration.
Otherwise, the module settings override any settings in the `config/ColdBox.cfc`

Examples:

```cfc
// ModuleConfig.cfc
component {
    this.modelNamespace = "cbsvg"; // defaults to module name
    this.parseParentSettings = true; // default value

    function configure() {
        settings = {
            inline = false
            class = "icon"
        };
    }
}

// config/ColdBox.cfc
component {
    function configure() {
        moduleSettings = {
            cbsvg = {
                inline = true
            }
        };
    }
}

// Merged settings in both
cbsvg = {
    inline = true,
    class = "icon"
};
```

```cfc
// ModuleConfig.cfc
component {
    this.modelNamespace = "cbsvg"; // defaults to module name
    this.parseParentSettings = false;

    function configure() {
        settings = {
            inline = false
            class = "icon"
        };
    }
}

// config/ColdBox.cfc
component {
    function configure() {
        moduleSettings = {
            cbsvg = {
                inline = true
            }
        };
    }
}

// Merged settings in both
cbsvg = {
    inline = false,
    class = "icon"
};
```

These settings can be accessed either through the module settings or the ColdBox settings:
```cfc
// Inject all module settings
property name="settings" inject="coldbox:modulesettings:{module}";
// Inject all module settings from the app settings
property name="settings" inject="coldbox:setting:{module}";
// Inject a single module setting
property name="customSetting" inject="coldbox:setting:customSetting@{module}";
```

### DIFF
--- a/system/web/services/ModuleService.cfc
+++ b/system/web/services/ModuleService.cfc
@@ -228,6 +228,10 @@ I oversee and manage ColdBox modules
 					disabled			= false,
 					// flag that says if this module can be activated or not
 					activate			= true,
+					// flag that determines if the module settings overrides any
+					// module settings in the parent config (ColdBox.cfc) or
+					// if the parent settings get merged (and overwrite the defaults).
+					parseParentSettings = true,
 					// Module Configurations
 					path				 	= modLocation,
 					invocationPath 			= modulesInvocationPath & "." & modName,
@@ -700,11 +704,23 @@ I oversee and manage ColdBox modules
 			if( structKeyExists( oConfig,"activate" ) ){
 				mConfig.activate = oConfig.activate;
 			}
+			// Merge the settings with the parent module settings 
+			if( structKeyExists( oConfig, "parseParentSettings" ) ){
+				mConfig.parseParentSettings = oConfig.parseParentSettings;
+			}
 
 			//Get the parent settings
 			mConfig.parentSettings = oConfig.getPropertyMixin( "parentSettings", "variables", {} );
 			//Get the module settings
 			mConfig.settings = oConfig.getPropertyMixin( "settings", "variables", {} );
+			// Add the module settings to the parent settings under the modules namespace
+			if ( mConfig.parseParentSettings ) {
+				// Merge the parent module settings into module settings
+				var coldboxConfigModuleSettings = controller.getSetting( "ColdBoxConfig" )
+					.getPropertyMixin( mConfig.modelNamespace, "variables", structnew() );
+				structAppend( mConfig.settings, coldboxConfigModuleSettings, true );
+			}
+			appSettings[ mConfig.modelNamespace ] = mConfig.settings;
 			//Get module datasources
 			mConfig.datasources = oConfig.getPropertyMixin( "datasources", "variables", {} );
 			//Get Interceptors
--- a/system/web/services/ModuleService.cfc
+++ b/system/web/services/ModuleService.cfc
@@ -716,9 +716,16 @@ I oversee and manage ColdBox modules
 			// Add the module settings to the parent settings under the modules namespace
 			if ( mConfig.parseParentSettings ) {
 				// Merge the parent module settings into module settings
-				var coldboxConfigModuleSettings = controller.getSetting( "ColdBoxConfig" )
-					.getPropertyMixin( mConfig.modelNamespace, "variables", structnew() );
-				structAppend( mConfig.settings, coldboxConfigModuleSettings, true );
+				var parentModuleSettings = controller.getSetting( "ColdBoxConfig" )
+					.getPropertyMixin( "moduleSettings", "variables", structnew() );
+				if ( ! structKeyExists( parentModuleSettings, mConfig.modelNamespace ) ) {
+					parentModuleSettings[ mConfig.modelNamespace ] = {};
+				}
+				structAppend(
+					mConfig.settings,
+					parentModuleSettings[ mConfig.modelNamespace ],
+					true
+				);
 			}
 			appSettings[ mConfig.modelNamespace ] = mConfig.settings;
 			//Get module datasources


bdw429s:
What about putting them into a struct called `modules.modulename` in the Coldbox config?  That way they're grouped a bit better and less likely to overlap with another variable in the Coldbox config.

``` js
component {
  function configure() {

    // Override module setting by convention
    moduleSettings = {
      // cbsvg module
      cbsvg = {
        setting1 = 'overrideValue',
        setting2 = 'overrideValue'
      },
      // cbi18n module
      cbi18n = {
        foo = true,
        bar = false
      }
    };

  }
}
```

elpete:
I'd be fine with that.

One question I did have is currently, if you choose to `parseParentSettings`, I'm updating the module's `settings` with the merged struct.  I can't decide if I think that is good or not.  Thoughts?

bdw429s:
I think that makes sense.

lmajano:
I don't personally like the idea of having a top level `moduleSettings`. We have always just used the name of the module as the convention. The only override might be with internal ColdBox settings, but since we have already several establishded. I am not sure if we need to go that route.

Also, it might seem that this might break some compatibility as well?

elpete:
Since we have never looked this up before, I'm not sure how it would break compatibility.

If a module is running it's own `parseParentSettings()` functions, it will still work just how it is implemented.

elpete:
To be fair, I'm torn on a top-level `moduleSettings` vs. `root`.  My arguments are style (`moduleSettings` is kind of ugly) vs. collisions (`root` could run in to other namespaces).

bdw429s:
@lmajano I'm aware of your preference here and I've always disagreed :)  Eric and I talked in Slack about this, but I'll reproduce some of the conversation here.

I think it made sense back in the ColdBox 3 days to have top level struct in the config for things like the debugger, validation, or i18n since we had very few cases of this, and they were well known so there was no chance of overlaps with other variables of functions on the page.  Now we need to consider people building apps with a large number of random 3rd party modules bearing unpredictable names.  In my opinion, we have to encapsulate these in some sort of top-level data structure.  In addition to helping prevent name collisions, it just makes good organizational sense.  

The second reason I think there should be a consistent struct for overriding module settings is because I'm thinking to the future when we (someday??) finally implement mutli-version module dependencies where two versions of the same module can be installed at once.  I see two additional places we'll want to be able to nest module config overrides.  One is in the `ModuleConfig.cfc` of the parent module, and the second is a nested syntax in the ColdBox app that can override settings in nested modules.  I think it will be much easier if we have some sort of consistent struct that always contains module setting overrides, regardless of what "level" you're overriding or whether you're in the Coldbox config or Module config.

> We have always just used the name of the module as the convention.

You may have done this personally, but I never did.  In the modules I wrote that needed parent setting overrides, I used different conventions.    Here's an example where I had the parent settings go in the settings struct:
https://github.com/bdw429s/Weather-Lookup-By-IP/blob/master/ModuleConfig.cfc
It's also worth noting that I've already introduced a documented standard into CommandBox where module settings are overridden by convention like what I suggested to Eric above.
The only reason there appears to be a precedent of any kind in ColdBox is simply because we put off this conversation for 2 years and your modules happen to follow a pattern you like :)

>  break some compatibility as well?

Not at all.  Modules can switch over to use the convention at any time.

lmajano:
I have evaluated your points.  I would say then **yes** to a top level `moduleSettings` organization bucket in order to avoid confusion for later versions.

bdw429s:
``` 
                                          `::..   .
                           `?XXX.  `T{/:.   %X/!!x "?x.
                             "4{7@( '!+!!X(:.`4!!X!x.?h7h
                         `!(:. ~!!!f(~!!!+!!{{.'~+h!tX!!?hh:.
                    '`X!.  !(d!X!!H!?{{``"!:?{{!{X*!?tX!!H*))h.
                  ...  '!X(!X!{{?@f!!!{!{x.!!%!!!%!!!)@Thh!!X)!).
                   ^!!!{:!(((!!: ~((({!!!h+!{{!X!+%?+{!!?!+)!+X(!+
               -    `\tXX{(~!!!!!:.!.%%(!!!!!!!!!X!))!!!!X%``%!!!(>
               ^X>:x. {!!!!X: ~!!*!{!!!{!~!X!)%!{!!!)?@!!!?!)?!!!>~
                 `X(!!:!!!{{(!!.)!%(:\!!:%~!~\!t!! `H!)~~!!!!!!(?@
                  `!X: `)!!!C44XX!!!.%%.X:>-> %!!X! /!~!.'!> !S!!!
              +{..  \X%\.'{??X!!!t!!~!!{!~!~'.!~~~ -~` {> !~ /!X`
                `X!XXM!!4!%\(4!!!!%(`,zccccd$$$$$$$$$ccx ` .~
                  "XLS@!)!!%L44X!!! d$$$$$$$$$$$$$$$$$$$,  '^
                   `!X?%:!!??X!4?*';$$$$$$$$$$$$$$$$$$$$$
                  `iXM:!!?Xt!XH!!! 9$$$$$$$$$$$$$$$$$$$$$
                   `X3tiXS#?WH!X!! $$$$$$$$$$$$$$$$$$$$$$
                   .MX?*StXX?X!!W? $$$$$$$>?$$$$$$$$$$$$
                    8??M%T%' r `  ;$$$$$$$$$,?$$$$$$$$$F
                    'StMX!': J$$d$$$$$$$$$$$$h ?$$$$$$"
                     tM9MH d$$$$$$$$$$$$$$C???r{$$$F,r
                     4M?t':$$$$$$$$$$$$$$$$h. $$$,cP"
                     'M>.d$$$$$$$$$$$$$$$$$>d$$${.,
                      ,d$$$$$$$$$$$$$$$$$'cd$$$$r"
                      `$$$$$$$$$$$$$$$??$Jcii`?$h
                       $$$$$$$$$$$$$F,;;, "?h,`$$h
                      j$$$$$$$$$$$$$.CC>>>>c`"  `"        ..,g q,
                   .'!$$$$$$$$$$$$$' `''''''            aq`?g`$.Bk
               ,- '  "?$$$$$$$$$$$$$$d$$$$$c, .         .)od$$$$$$
          , -'           `""'   `"?$$$$$$$$$??=      .d$$$$$$$$F'
        ,'                           `??$$P       .ed$$P""   `
       ,                                `.      z$$$"
       `:dbe,                          x,/    e$$F'
       :$$$$P'`>                       $F  z$$$"
      d$$$P"'  >                       $Fe$$$"
    .$$$?F     ;                       $$$$"
    $$$$$$eeu. >                       >P"
     `""???$$$$$eu,._''uWb,            )
               `""??$P$$$$$$b.         :
                >     ?$$$"'           {
                F      `"              `:
                >                       `>
                >                        ?
               J                          :
               X                ..  .     ?
               "{ 4{!~;/'!>{`~{>~.>! ~! '"
                '>!>=.%=.;~~>~4~`{'>>>~!
                 4'!/>!\\!{~~:/{;!{;`;/=':
                 `=;!~:`~!>{.-; "(>=.':!;'
                  :;=.~{`;`~>!~> ?!/>>~!!{'
                  ~:~'!!;`;`~:>); ;(.uJL!~~
                    >L.(.:,L;L:-+d$$$$$$
                    :4$$$$$$$L   ?$$#$$$>
                     '$$$B$$$>    $$$MB$&
                      $$$$$$$      $$$@$F
                      `$$$$$$>     R$$$$
                       $$$$$$     {$$@$P
                       $R$$$R     `!)=!>
                       $$$6T       $$$$'
                       $$R$B      ;$$$F,._
                       !=!(!    .'        ``= .
                       $$$$F    (.             '\
                     ,{$$$$(      ``~'`` --:.._.,)
                    ;   ``  `-.
                    (          "\.
                     ` -{._       ".
                           `~:,._ .:
```

elpete:
Merge this whenever, folks. 😄 

I'll go update the docs.

PR #291, opened 2016-10-20
Provide a framework convention for loading env secrets

elpete:
This pull request makes it easier to use secrets stored in the environment.  It does this via two methods:
1. It provides `java.lang.System` as a property injected into the `config/ColdBox.cfc` config file and any module's `ModuleConfig.cfc` file.  This saves the framework user from having to manually instantiate it.
2. It defaults to copying over env keys and values to Java properties.  This can be opted out by specifying `this.copyEnvToProperties = false` on the config file.  The reasons it copies env secrets over to properties is two-fold:
   a. All potential secrets are available in the Java properties instead of having to potentially check two locations.
   b. The `java.lang.System.getProperty` method allows for a second argument of `defaultValue` where `java.lang.System.getEnv` does not.

Things this pull request does not do:
1. It does not set environment secrets.  That must be done either manually or using a tool like [commandbox-dotenv](https://www.forgebox.io/view/commandbox-dotenv).
2. It does not load the Java properties in to the ColdBox settings.  This could be added if desired.

Thoughts on conventions are welcome, as are comments on the code. 😄

### DIFF
--- a/system/web/config/ApplicationLoader.cfc
+++ b/system/web/config/ApplicationLoader.cfc
@@ -87,6 +87,7 @@ Loads a coldbox cfc configuration file
 		oConfig.injectPropertyMixin( "controller",instance.controller);
 		oConfig.injectPropertyMixin( "logBoxConfig",instance.controller.getLogBox().getConfig());
 		oConfig.injectPropertyMixin( "appMapping",configStruct.appMapping);
+		oConfig.injectPropertyMixin( "system", createObject( "java", "java.lang.System" ) );
 
 		//Configure it
 		oConfig.configure();
--- a/system/web/config/ApplicationLoader.cfc
+++ b/system/web/config/ApplicationLoader.cfc
@@ -89,6 +89,10 @@ Loads a coldbox cfc configuration file
 		oConfig.injectPropertyMixin( "appMapping",configStruct.appMapping);
 		oConfig.injectPropertyMixin( "system", createObject( "java", "java.lang.System" ) );
 
+		if ( shouldCopyEnvToProperties( oConfig ) ) {
+			copyEnvToProperties();
+		}
+
 		//Configure it
 		oConfig.configure();
 
@@ -860,6 +864,23 @@ Loads a coldbox cfc configuration file
 		<cfscript>
 			return reReplace(arguments.appMapping,"(/|\\)",".","all" );
 		</cfscript>
+    </cffunction>
+
+    <!--- shouldCopyEnvToProperties --->
+    <cffunction name="shouldCopyEnvToProperties" output="false" access="private" returntype="boolean" hint="Returns true if env keys and values should be transferred to Java properties">
+    	<cfargument name="oConfig" type="any" required="true" />
+		<cfreturn ( NOT structKeyExists( oConfig, "copyEnvToProperties" ) OR oConfig.copyEnvToProperties EQ true ) />
+    </cffunction>
+
+    <!--- copyEnvToProperties --->
+    <cffunction name="copyEnvToProperties" output="false" access="private" returntype="void" hint="Copies env keys and values to Java properties">
+    	<cfscript>
+    		var system = createObject( "java", "java.lang.System" );
+			var envs = system.getEnv();
+			for ( var key in envs ) {
+			    system.setProperty( key, envs[ key ] );
+			}
+    	</cfscript>
     </cffunction>
 
 </cfcomponent>
--- a/system/web/services/ModuleService.cfc
+++ b/system/web/services/ModuleService.cfc
@@ -634,6 +634,7 @@ I oversee and manage ColdBox modules
 			oConfig.injectPropertyMixin( "wirebox", 		controller.getWireBox() );
 			oConfig.injectPropertyMixin( "binder", 			controller.getWireBox().getBinder() );
 			oConfig.injectPropertyMixin( "cachebox", 		controller.getCacheBox() );
+			oConfig.injectPropertyMixin( "system",          createObject( "java", "java.lang.System" ) );
 
 			// Configure the module
 			oConfig.configure();


elpete:
I'm not sure about this failure reason — `can't create directory /logs`.  Can @lmajano or @bdw429s look in to it, please?

elpete:
Another option I have been thinking about is a custom `env( key, default )` that does a similar thing, but could even look in both areas (instead of copying them over). Or a combination of any of these.

bdw429s:
Just some thoughts on this-- Lucee allows you to use placeholders like `${my.env.variable}` as a replacement in config files.  I think that's mainly needed so you can reference properties in XML files where there's no interpolation of CFML variables.  I'm thinking Luce 5 also has a special scope for environment vars, but that's not of much use to users on Lucee 4 or Adobe CF.

I like the idea of this and I think it will be more and more needed as people move to CommandBox or cloud solutions where they spin  up servers ad-hoc.  While accessing the variables easily from your config is important, I'm kind of interested in some sort of naming convention that would automatically get picked up by ColdBox to set/override settings.  I'd say to put this in a module, but I'm not sure modules load soon enough in the ColdBox application lifecycle for that to actually work.

If we do it, I'd certainly want to support struct or array notation whereby a env variable or java property named something like
```
coldbox_setting.layouts.defaultLayout=MyLayout.cfm
```
would override the correct key in the actual config.   I have a lot of work in CommandBox's JSONService to help with parsing of such strings which brings me to the next idea which is auto-expanding JSON strings in env vars or system properties for them to contains complex values like so:
```
coldbox_settings.interceptors.ses="{ class='coldbox.system.interceptors.SES' }"
```
I support this for settings config/server/package properties via the CLI.  Then you have to deal with things like appending to the struct/array vs overwriting but I think it's food for thought to create some very powerful ways to override config at runtime based on the environment.

elpete:
>  I'd say to put this in a module, but I'm not sure modules load soon enough in the ColdBox application lifecycle for that to actually work.

This is true.  I tried a module at first, but the config has already been loaded.  Reloading it seemed strange.

> If we do it, I'd certainly want to support struct or array notation whereby a env variable or java property named something like
```
coldbox_setting.layouts.defaultLayout=MyLayout.cfm
```
> would override the correct key in the actual config.

Interesting.  I based my pull request loosely off of [Laravel's implementation](https://laravel.com/docs/5.3/configuration#environment-configuration).  Another project that follows this is [dotenv](https://github.com/bkeepers/dotenv).  In each case, it seems to be a simple `.properties` like file, not nested.  I think the mindset I was approaching from was that you can choose which environment variables to use, if any, and in what settings.  There surely would be nothing stopping you from setting a key of `coldbox_setting.layouts.defaultLayout` and a value of `MyLayout.cfm`, but I think you would have the flexibility in the `config/ColdBox.cfc` to assign that value to `settings = { foo = system.getProperty( "coldbox_setting.layouts.defaultLayout" ) };` if you wanted to.  Said a different way, I'm not sure I want all the environment variables inserting themselves in to my config — just the ones I choose.

So, I think I am in support of the user implementing which environment variables become ColdBox settings rather than automatic settings based on key paths.  My two cents.

bdw429s:
>  I'm not sure I want all the environment variables inserting themselves in to my config — just the ones I choose.
Heh, well who's creating the env vars?  I assume that's you!  So it's not like they become sentient and start taking over your application :)

Either way, it was just a thought off the top of my head to see if there was some generic way to basically set any setting via a convention and have it just get sucked in.   My main issue with the other way (Simply providing the values and it's up to the app to use them) was that it has some serious drawbacks in Lucee for instance.  It's easy enough to set your database password in an env var and use it in the config, but what if you want one environment to have 3 completely new datasources?  You can't really control that via just simple values that require you to use them.  That's what got me thinking about making it a bit "smarter" if you will.

Now, even if we don't create any "auto-injected naming convention" and we leave it to the user to use the variables as they wish, it's still possible for us to do auto-expansion of JSON.  We actually used to do this back in the olden days when ColdBox loaded its config from XML files.  Any JSON strings were automatically deserialized.

elpete:
Interesting points.  I guess I just assumed that if you wanted three different datasources you'd have something like this:

```
ORACLE_HOST=10.0.0.1
ORACLE_PORT=1521
ORACLE_USERNAME=admin
ORACLE_PASSWORD=password

MYSQL_HOST=10.0.0.1
MYSQL_PORT=3306
MYSQL_USERNAME=admin
MYSQL_PASSWORD=password

OTHER_MYSQL_HOST=otherdomain.net
OTHER_MYSQL_PORT=3306
OTHER_MYSQL_USERNAME=admin
OTHER_MYSQL_PASSWORD=password
```

In any case, those are some interesting ideas, especially the auto-expanding json.  Also, it wouldn't force anyone to use it.  People like could could stick to our verbose, one-string-at-a-time ways. :wink:

bdw429s:
@lmajano have you digested yet, or do you just want Eric and I to finish hammering it or so this didn't drag on?

lmajano:
I am still digesting. But by all means discuss. It seems we will need to get into environment variables due to containers. So having an official way to retrieve and access them is necessary. I am just not sure of mixing them with settings. Unless we want to alter any setting via  env variable

jclausen:
Mind if I weigh in, since I was auto-subscribed to the thread? :)  Why not just add the Java-style placeholder (`${ENVIRONMENT_VARIABLE}`) for environmental variables as part of Coldbox.cfc settings load and evaluate them there.  As with using them in Bash/Java, you'd want to return the server engine null equivalent (e.g. empty string or true null) if they aren't available.  You'll add some recursion to the settings load in `system.web.config.ApplicationLoader`, though.  I agree, as well, that I wouldn't want to have all environment vars injected in to the configuration.  That's overkill, IMHO.  

For container deployments with Lucee, I regularly use them in my datasource creation, since I can specify the db user/pass as env variables and then grab them in App.cfc.  That really wouldn't be covered, anyway, by a Coldbox config setting and it would still be necessary to use `java.lang.System`.    That being said, this is probably more of a Lucee roadmap issue than it is a framework problem to be solved.  Two of the TAG members happen to be subscribed to this thread! [hint, hint, wink]

It would be handy to have them easily available through a supertype `getEnv( name, defaultValue)` supertype method. Again, though, the place any environment variables should probably be dealt with is on App start or Coldbox load.

bdw429s:
@elpete Just reading back through the conversation again today.  I realized we may have been talking about different things earlier.  Anyway, a few more thoughts:

1. The idea of a convention to override settings wouldn't be mutually exclusive with the ability to just define ad-hoc env vars and access them.  The user would opt-in by choosing to prefix an env var with a special string with as `coldbox_setting.settingname`.
2. You mentioned it would be more flexible to choose in your config where to use the env var.  I would actually say it's the opposite.  If a user opted to use the auto-inserting convention, their env vars would be treated as an automatic override that wouldn't allow any code at all in the ColdBox.cfc.  In fact, the ColdBox config wouldn't even need to know if a setting was being overridden.  This would actually allow the application to still work on default values in the absense of any env vars at all.
3. Regarding the example of overriding data sources-- I don't think I explained it quite well enough.  Imagine you used MySQL on dev and Oracle on production and the struct of properties for each of those was completely different meaning the Oracle data source required additional settings that the MySQL one didn't and vice versa.  You can't control the _keys_ being set with simple placeholders, you can only control the _values_.  That's why I was brainstorming ways to, say, set an entire struct in the config to be 100% unique per environment instead of just being able to control a set number of values of keys in a struct.  

So all that aside, (I'm not necessarily sold on it, just wanted to make sure we were on the same page).   I actually agree with your later idea of providing our own CFML-based function to access these value instead of exposing the Java classes themselves.  This allows us to do many things such as mock them, etc and ensures we have control over the API.

bdw429s:
@jclausen Thanks for weighing in with more thoughts!  Here's my resultant thoughts, though I'm beginningn to think a RoadBox meeting would be an ideal way to hash some of this out...

>  Why not just add the Java-style placeholder (${ENVIRONMENT_VARIABLE}) 

Yep, I suggested that too in my first reply.  I think the answer of why we don't need that though, is because the ColdBox config is already written in CFML and can simply interpolate CFML variables.  We don't need to build another system for interpolating variables like we would if this were XML or something.

>  I wouldn't want to have all environment vars injected in to the configuration. That's overkill, IMHO.

To be clear, unless I misunderstood what Eric has said, I don't think anyone has advocated for insertion of all env vars into the config!  You may have gotten that from my suggestion of having a convention by which specific, selected env vars would automatically override config settings based on their name.  So given any env var who's name started with `coldbox_setting.` we'd parse the next string and set it's value into the variables scope of the config CFC.  That means you could have 1000 env vars and they'd all be completely ignored, but if you had a single one called:
```
coldbox_settings.foo=bar
```
Then we'd override the `foo` setting in your config to equal `bar`.  I was thinking long the same lines that your ColdBox config can now override a module's settings without the module needing to put any plumbing in place to allow for it (or expect it).\

> have them easily available through a supertype 
It's worth noting the ColdBox config doesn't extend the framework super type so we'd need to inject those helpers manually into there.  I had also thought about adding something to the actual framework super type as you suggested though it wouldn't be any different than dealing with env vars in the config like you suggested and then using the values as settings like you suggested.

jclausen:
@bdw429s The Java-style placeholder would really only be useful in the sense that the supertype methods aren't available at that point.  It would be great to have a CFML function available for that but, unless you used an implicit scope (blech), or assigned an unscoped variable to that function double-blech), I can't see how you'd elegantly do that so early in the application load.

I see what you're saying on the setting.  It gets a bit tricky, though, when you're dealing with module settings.  Take for example `cbMongoDB`.  In order to use a password secret for connections, the environment variable would have to use some sort of pseudo array syntax, since the password for the host is in the first host entry in the array of servers:  `coldbox_setting.MongoDB.hosts.1.password=myPassword`

Other than for top-level settings values, it could get ugly to handle, which is kind of why I think it's something the Lucee team (and Adobe team) need to discuss adding a core function for this purpose.

bdw429s:
>  It would be great to have a CFML function available for that but, ... I can't see how you'd elegantly do that so early in the application load.

Maybe I'm missing something, but I assumed we'd inject a helper UDF the same way everything else is injected into the config right here in the application loader:
https://github.com/ColdBox/coldbox-platform/blob/development/system/web/config/ApplicationLoader.cfc#L87-L89

> environment variable would have to use some sort of pseudo array syntax,

Actually, they could just use the standard array syntax :)  I already do this in CommandBox right now with the `server set`, `package set`, and `config set` commands.
```package set myArray[2]=value```
or
```package set key.nestedKey.nestedAgain[42].anotherKey=value```
The logic for all this is in CommandBox's `JSONService`.

**Comment by jclausen** [comment]
> Actually, they could just use the standard array syntax :) 

No, I meant using that when declaring the environment variable, since you can't use arrays there. :)

> Maybe I'm missing something, but I assumed we'd inject a helper UDF the same way everything else is injected into the config right here in the application loader

Ah, I wasn't aware that those property mixins were injected there.  That's a great way to do it, then.

**Comment by bdw429s** [comment]
@jclausen Aw, man that sucks that bash doesn't allow that.  Windows actually has no issues with names like that.
```
set foo.bar=baz
set bum[1]=blargh
echo %foo.bar%
echo %bum[1]%
```
Well, I guess that sort of rules out creative names of env vars that parse to match CFML data structures-- at least in env vars anyway.

**Comment by lmajano** [comment]
ok, this thread got out of hand and too long of a novel to read.  Can we summarize if this is needed?

**Comment by jclausen** [comment]
@lmajano : I would vote for a property mixin `getEnv` method that can be used from within the coldbox config.

**Comment by bdw429s** [comment]
I think this pull is fine and just needs the addition of a helper method.  We never decided on a name nor if we'd have one method for env vars and another for java properties, or a single one that looks in both places.  The rest of the "novel" above was discussing possible future development that this pull might lead to.  I think it needs a RoadBox discussion to finish hammering out
.

**Comment by elpete** [comment]
Ideally, we'd all use `server.settings` like Lucee 5.

In the meantime, I propose 3 helpers:
```cfc
function getEnv( key, default );
function getProperty( key, default );
function getSecret( key, default ); // checks env first, then properties, then uses default or fails
```
If we like that, I can modify the pull so it's not copying all the env variables over to the system properties as well.

bdw429s:
I don't know about `getSecret()` since not everything there might be "secret" per se, but I can't think of a better name so I don't really care :)  If it's a standard in other languages/tools then that's fine.  I think the remaining question is whether these methods should get added to the framework super type or just the config.  Just the config for now is fine.  I'm adding a link to this to our Roadbox agenda so we can discuss there.

lmajano:
I want to start revising this again as we move more to containers. 

Let's start with that yes we need a way to read environment variables and system properties. 

What if we have a single entry point menthod where properties take precedence then environment. 

GetEnvironment( setting, default )

A part from a supetype helper we also need a wire box dsl to allow for injection of individual items or maybe the entire environment.

bdw429s:
@lmajano I think we were going to discuss this in RoadBox, but you and @elpete haven't been able to attend at the same time yet (or we've not had the meeting).

elpete:
To be fair, I can't attend RoadBox since it's in the middle of the work day for me.

lmajano:
Well, I just dont' want to leave this hanging. So wanted to get a sense of were to take this.

elpete:
I've taking our discussion here and created a new pull request for it over here: https://github.com/ColdBox/coldbox-platform/pull/305

I'm going to close this pull request in favor of the new one.

PR #296, opened 2016-11-02
Run Travis for multiple cfengines

elpete:


elpete:
I'm not super familiar with Ant, so I may need help fixing the build errors (mostly on Adobe engines).

Also, it looks like the Lucee 5 error is due to the cache not being defined.  I believe that's because the XML file is stored in a different place for Lucee 5.  That will need to be updated in the build as well, but I'm hazy on the details.

lmajano:
I think the issue is the passing of the cfengine variable:
```
     [exec] Error getting ForgeBox entry [$ENGINE]  The entry slug sent is invalid
```

elpete:
I fixed that error with my second commit.

.travis.yml
```diff
@@ -37,7 +42,7 @@ install:
   # Store the Ortus API Key for forgebox publishing in `master`
   - box config set endpoints.forgebox.APIToken=$FORGEBOX_API_TOKEN
   # startup server so we can get directory location first
-  - box server start
+  - box server start cfengine=$ENGINE
```
lmajano: how about adding this to this line: `name=$ENGINE` so it doesn't collide.

elpete: I like it.

lmajano:
seems like all adobe engines fail

bdw429s:
Does Travis have access to the internet? There's a bug in Adobe engines right now on CommandBox stable where they won't start without internet connectivity

elpete:
Implemented in https://github.com/ColdBox/coldbox-platform/commit/5898465da9e75312d54e0a54194014a7861ab62f

lmajano:
Basing it off of yours, but tweaking a few things. thanks @elpete

PR #298, opened 2016-11-08
Add postReinit interception point

elpete:
Add a postReinit interception point that fires after the `loadColdBox` method finishes.

My specific use case for this is to have an interceptor call `ormReload` on a `fwreinit`, but only after the configuration file (`config/ColdBox.cfc`) has been loaded.  This is because I am using the `cfapplication` tag to update my default datasource based on the environment (`cfapplication( action="update", datasource="my_env_specific_datasource" );`)  Calling `ormReload` in the `preReinit` interception point is too early as the datasource has not been updated.

There are likely other use cases as well.  Looking forward to your feedback.

### DIFF
--- a/system/Bootstrap.cfc
+++ b/system/Bootstrap.cfc
@@ -104,6 +104,11 @@ component serializable="false" accessors="true"{
 
 					// Reload ColdBox
 					loadColdBox();
+					// If we are Reiniting, fire the postReinit interception point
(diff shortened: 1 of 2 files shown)


bdw429s:
@elpete out of curiosity, can you use afterConfigurationLoad?  I think the only difference is that it also fires I first app load.

elpete:
That's the perfect interception point!  Thanks!

PR #299, opened 2016-12-06
Fix bug where modules would override the settings struct of config/ColdBox.cfc

elpete:
This bug was introduced when adding the parse parent settings by convention for modules (https://github.com/ColdBox/coldbox-platform/pull/289).
Instead of updating the module namespace in the `moduleSettings` struct to match, it updated the module namespace in the `settings` struct.
Some modules depended on the old behavior, especially those that rolled their own parse parent settings.

This pull request fixes the behavior so that the module's settings and the module namespace in `moduleSettings` are kept in sync.  `settings` is not touched.

### DIFF
--- a/system/web/services/ModuleService.cfc
+++ b/system/web/services/ModuleService.cfc
@@ -727,7 +727,7 @@ I oversee and manage ColdBox modules
 					true
 				);
 			}
-			appSettings[ mConfig.modelNamespace ] = mConfig.settings;
+			appSettings.moduleSettings[ mConfig.modelNamespace ] = mConfig.settings;
 			//Get module datasources
 			mConfig.datasources = oConfig.getPropertyMixin( "datasources", "variables", {} );
 			//Get Interceptors
--- a/test-harness/config/Coldbox.cfc
+++ b/test-harness/config/Coldbox.cfc
@@ -48,7 +48,9 @@
 
 		// custom settings
 		settings = {
-
+			test1 = {
+				display = "not-core"
+			}
 		};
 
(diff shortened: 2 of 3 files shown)


lmajano:
builds are failing @elpete

bdw429s:
@lmajano The build failed on your config-copying recipe:
```
can't copy file [build/lib/lucee-server.xml] to [/home/travis/.CommandBox/engine/cfml/server/lucee-server/context]
No such file or directory
```
Doesn't appear to be related to @elpete 's code.

bdw429s:
@lmajano I figured out the issue.  Your Travis build is doing this:
```start cfengine=lucee@4.5```
Which is basically asking CommandBox to start the latest stable version of CommandBox starting with 4.5.  When you set up the build, `4.5.3.020` was the latest version of ForgeBox as well as the version of Lucee that CommandBox `3.4.0` runs on.  Last week I added Lucee `4.5.4.017` to ForgeBox (which CommandBox `3.4.1-snapshot` runs on) which now means the build is downloading the `4.5.4.017` WAR and doing a WAR server start, so the internal server is not being used and that means the path that the config needs to be run is different.  
The quick fix is to leave off the version entirely from the start command when you want Lucee 4.x and it will use the internal version again, but that's not a very good long term solutions since eventually CommandBox will run on Lucee 5 and so on.  I think the long term fix is we add some better plumbing into CommandBox to get the web/server context of servers and use it to copy the configs.

bdw429s:
@elpete I put in a temp fix on the `development` branch to fix the build for now.  Can you rebase your branch please?

elpete:
Rebased and passing

PR #305, opened 2017-02-08
Provide helper methods for system settings that covers both properties (first) and env variables.

elpete:
This pull request is the spiritual successor to https://github.com/ColdBox/coldbox-platform/pull/291

cc @lmajano @bdw429s @jclausen 

Usage:

+ `getSystemSetting( key, defaultValue )` -> looks in properties first, env second. Returns the `defaultValue` if neither exist. 
+ `getSystemProperty( key, defaultValue )` -> Returns the Java System property.  Returns the `defaultValue` if does not exist.
+ `getEnv( key, defaultValue )` -> Returns the Env property.  Returns the `defaultValue` if does not exist.

These helper methods are available in the `config/ColdBox.cfc` file and all `ModuleConfig.cfc` files.

### DIFF
--- a/system/core/util/Util.cfc
+++ b/system/core/util/Util.cfc
@@ -146,6 +146,83 @@ Description :
 
     </cffunction>
 
+    <!--- getSystemSetting --->
+    <cffunction name="getSystemSetting" output="false" access="public" returntype="any" hint="Retrieve a Java System property or env value by name.">
+    	<cfargument name="key" required="true" type="string" hint="The name of the setting to look up."/>
+		<cfargument name="defaultValue" required="false" hint="The default value to use if the key does not exist in the system properties or the env" />
+
+		<cfscript>
+			var system = createObject( "java", "java.lang.System" );
+			
+			var value = system.getProperty( arguments.key );
+			if ( ! isNull( value ) ) {
+				return value;
+			}
+			
+			value = system.getEnv( arguments.key );
+			if ( ! isNull( value ) ) {
+				return value;
+			}
+
+			if ( ! isNull( arguments.defaultValue ) ) {
+				return arguments.defaultValue;
+			}
+
+			throw(
+				type = "SystemSettingNotFound",
+				message = "Could not find a Java System property or Env setting with key [#arguments.key#]."
+			);
+		</cfscript>
+    </cffunction>
+
+    <!--- getSystemProperty --->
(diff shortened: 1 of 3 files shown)


lmajano:
This looks great @elpete, do we have a ticket in Jira for this? Also, I think that the super type will need these methods as well. Also, would any model need any of these features?

bdw429s:
I agree, this is a great start.  @lmajano @elpete What if next we created a new `@coldbox` mapping to a util object (or put them in the existing one).  The super type could point to them and people would also have something to inject into their models if they wanted to access there?

elpete:
I would think you would just set the secrets you cared about in your ColdBox settings.

```cfc
component {
    function configure() {
        settings = {
             environment = getEnv( "env" ),
             sendGridApiKey = getSystemSetting( "SEND_GRID_PRIVATE_KEY" )
        }
    }
}
```

Then you could access them like any normal settings:

```cfc
component {

    property name="apiKey" inject="coldbox:setting:sendGridApiKey";


    function index( event, rc, prc ) {
        var env = getSetting( "environment" );
    }
}
```

The idea of a new `coldbox` mapping is also interesting.  `coldbox:systemSetting:env"?

bdw429s:
Ooh, well a new injection DSL namespace would be cool too.  I was actually just thinking of a WireBox mapping to a utility that had the methods for models to use:
```js
property name="envPropUtil" inject="EnvPropUtil@coldbox";
```
but I like the idea of this as well.
```js
property name="myProperty" inject="coldbox:envProp:myProperty";
```
I just don't think we could handle defaults that way.

elpete:
I don't think we need to handle defaults with the DSL.  We don't for `coldbox:setting:{setting}`.

I can add this to the super type if the rest agree, but I don't think it's necessary.  I think the pattern would be to map the desired system settings to ColdBox properties in the `config/ColdBox.cfc` file.

elpete:
_Technically_, you could always inject the controller and get the util to get the new methods, but that would definitely look ugly. ;-)

lmajano:
@elpete I think this pull is ready to go. All I need is a Jira ticket for it and some docs on this.

elpete:
Jira ticket: https://ortussolutions.atlassian.net/browse/COLDBOX-572

elpete:
Docs PR: https://github.com/ColdBox/coldbox-docs/pull/40

lmajano:
Ok @elpete merged, but I forgot to check for tests. We need some tests here.

elpete:
Pull request here: https://github.com/ColdBox/coldbox-platform/pull/308

PR #309, opened 2017-03-20
var scope

sanaullah:
added missing var scope


lmajano:
Thanks @sanaullah I added manually to test some travis failure issues

PR #311, opened 2017-04-02
Adobe Coldfusion syntax corrections and CF-11 bug fixes.

sanaullah:
Adobe Coldfusion syntax corrections and CF-11 bug fixes.

Removed null support as its cause more problem in CF-11

CF-11 inside the cfc the include .cfm template can't access the methods which are only declare as property... (hack) have to create setter/getter methods

### DIFF
--- a/system/FrameworkSupertype.cfc
+++ b/system/FrameworkSupertype.cfc
@@ -144,7 +144,8 @@ component serializable="false" accessors="true"{
 		numeric collectionStartRow="1",
 		numeric collectionMaxRows=0,
 		collectionDelim="",
-		boolean prePostExempt=false
+		boolean prePostExempt=false,
+		name=""
 	){
 		return controller.getRenderer().renderView( argumentCollection=arguments );
 	}
--- a/system/FrameworkSupertype.cfc
+++ b/system/FrameworkSupertype.cfc
@@ -144,8 +144,7 @@ component serializable="false" accessors="true"{
 		numeric collectionStartRow="1",
 		numeric collectionMaxRows=0,
 		collectionDelim="",
-		boolean prePostExempt=false,
-		name=""
+		boolean prePostExempt=false
 	){
 		return controller.getRenderer().renderView( argumentCollection=arguments );
 	}
--- a/system/cache/providers/CacheBoxProvider.cfc
+++ b/system/cache/providers/CacheBoxProvider.cfc
@@ -40,11 +40,11 @@ Properties
 			// Eviction Policy
 			instance.evictionPolicy = "";
 			// Element Cleaner Helper
-			instance.elementCleaner		= CreateObject("component","coldbox.system.cache.util.ElementCleaner").init(this);
+			instance.elementCleaner		= createObject("component","coldbox.system.cache.util.ElementCleaner").init(this);
 			// Utilities
 			instance.utility			= createObject("component","coldbox.system.core.util.Util");
 			// UUID Helper
-			instance.uuidHelper			= createobject("java", "java.util.UUID");
+			instance.uuidHelper			= createObject("java", "java.util.UUID");
 
 			// CacheBox Provider Property Defaults
 			instance.DEFAULTS = {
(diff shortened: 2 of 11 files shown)


elpete:
I would like to find a way to keep the Null checks. Maybe a combination of `structKeyExists` and `isNull`? I think ripping out the null support is a bit hasty.

bdw429s:
Agreed.  What problems in CF11 were you having @sanaullah?  We worked hard to finally be compat with Lucee's full null support so we'd like to keep it.

elpete:
I think it's likely that there is a missing `isNull` check somewhere. `isNull` does the `structKeyExists` check and more.

sanaullah:
Hi Brad / Eric

@Eroc Really simple why not run your dev branch with CF-11 and test your IsNull implementation. 

@Brad ColdBox first priority to be compatible with Adobe CF, not the Lucee. You can create separate version for Lucee with full null support.

Once all fixed Let me know :) easy peasy

sanaullah:
Really simple test code, which cause issues in CF-11
http://trycf.com/

```
<cfscript>
ar = [];
t1 = {};

ar[1] = t1;

if(!IsNull(ar[1].value))
    writeoutput("struct key exists check with IsNull");
else
    writeoutput("struct key exists check with IsNull--false");
</cfscript>
```

lmajano:
If we can keep both then that's the best of both world.

So let's figure out a way to keep acf compatibility and null support

Luis Majano
CEO
Ortus Solutions, Corp
P/F: 1-888-557-8057

bdw429s:
That's a dumb Adobe bug.  I put in this ticket here for it: https://tracker.adobe.com/#/view/CF-4198479
I disagree that we shouldn't worry about Lucee and only support ACF.  Ortus pretty much exclusively uses Lucee for our servers and many of our users use it as well.  The full null support was added due to demand from the community for it.  

I think we can figure out a syntax that works for both.  In this case we can probably pull the struct out into an intermediate variable to get around the ACF bug.  We are still having issues getting Adobe to work with our Travis builds, but once we do, it will be easier to test this sort of stuff right away.  Actually @elpete, I was thinking, we need to find a way to Add Lucee with full null support to our Travis matrix to make sure we test ColdBox with and without it.

sanaullah:
@bdw429s We can wait for Adobe response and fix. Also I really don't like to add any intermediate variables to make it work. CB code should be neat, precise and fast.

elpete:
Unfortunately, Adobe's response time is...lackluster.  They might close the ticket for some inexplicable reason or even only fix it in ACF 2016+.  I think we need to add additional checks to maintain compatibility with ACF 10+ (9+?).

bdw429s:
@sanaullah @elpete Lol, I wasn't implying that we wait for Adobe to fix anything, but we have to put in a ticket so it will at least get fixed at some point in the future.  Stepping around engine-specifics is just part of writing a framework.  We used to have tons of code just for single versions of Adobe CF to work.  That's all been scaled down over the years as the engines became more compatible.  I'd take a couple extra lines over that any day.  And I'm not concerned about the performance of a single set statement.  It's just a matter of 
```js
var thisDIArg = DIArgs[ x ];
thisDIArg.javaCast
```
is as or more readable than 
```js
DIArgs[ x ].javaCast
```
which will actually make for simpler code since `DIArgs[ x ]` is used over and over.

@lmajano No, `strcutKeyExists()` will not work in Lucee when full null support is enabled.  That's because the values are being explicitly set to a `null` value which _does_ exist in Lucee's full null support.  That's one of the big differences.  A null struct value doesn't exist in ACF and "normal" lucee, but a null struct value does exist in Lucee with full null support (as it should).  That's why we use the `isNull()` check.

sanaullah:
@bdw429s
I never been in favor of creating more variables.

I do remember the time when I delete so many vars which just holds the values/object and do nothing else

Example
```
var oResult = xx.action();
return oResult;

Prefer way

return xx.action();
```

As your proposed new style to create more vars to deal with NULL issue.
Its simply without any doubt will effect the performance, so I will never be in favor of this.

Another example which breaks in CF-11 and works in Lucee
```
<cfscript>
t1 = [];

writeoutput(IsNull(t1[1]));

</cfscript>
```

bdw429s:
> Its simply without any doubt will effect the performance, so I will never be in favor of this
I've never in my life seen a variable creation affect real-world performance.  It doesn't even consume memory- it's just a reference on the heap.  Can you show me an example of this affecting performance?

Regarding `isNull()` with the bracket notation on Adobe, that is a known issue documented in these two tickets:
https://tracker.adobe.com/#/view/CF-409839
https://tracker.adobe.com/#/view/CF-3777274

PR #312, opened 2017-04-14
Adding an addResource method which creates the RESTful routes for a resource

murphydan:
Let me know if we're good on this and I'll get a pull request going for the docs as well. Thanks to @elpete for a lot of help on this one.

`addResource` is a method for quickly scaffolding the RESTful routes for a resource

Calling `addResource( "photos" )` would create the following routes: ([taken from the Rails Routing Guide](http://edgeguides.rubyonrails.org/routing.html))

HTTP Verb | Path | Handler.Action | Used For
----------- | -----| ---------------|----------------
GET | /photos | photos.index | Display a list of photos.
GET | /photos/new | photos.new | Return an HTML form for creating a new photo.
POST | /photos | photos.create | Create a new photo.
GET | /photos/:id | photos.show | Display a specific photo.
GET | /photos/:id/edit | photos.edit | Return an HTML form for editing a photo.
PATCH/PUT | /photos/:id | photos.update | Update a specific photo
DELETE | /photos/:id | photos.delete | Delete a specific photo

The `addResource` method has the following attributes:

* The first parameter `name` is required and is the name of the resource. This name will be used for the url path and handler name by default.
* The handler name can be overridden by passing it in as the second argument `handler`. By default, it is the same name as the resource name.
  * i.e. `addResource( name = "photos", handler = "PhotosController", only = "index" )` would generate the route `/photos` that runs the event `PhotosController.index`.
* The `id` placeholder can be overridden by passing in the desired name to the `parameterName` argument.
  * i.e. `addResource( name = "photos", parameterName = "photoId", only = "show" )` would generate the `/photos/:photoId` route.
* It can take an optional list or array of actions as the `only` parameter. If provided, these are the only actions that will be generated.
  * i.e. `addResource( name = "photos", only = "index,show" )` would generate `/photos` and `/photos/:id` routes.
* It can take an optional list or array of actions as the `except` parameter. If provided, these actions will not be generated.
  * i.e. `addResource( name = "photos", except = [ "new", "create", "edit", "update", "destroy" ] )` would generate `/photos` and `/photos/:id` routes.
* If both `only` and `except` are provided, both will be applied.
  * i.e. `addResource( name = "photos", only = "index,show", except="show" )` would generate the `/photos` route.

The tests are implemented in `tests.specs.interceptors.SESAddResourceTest`.

### DIFF
--- a/system/interceptors/SES.cfc
+++ b/system/interceptors/SES.cfc
@@ -393,6 +393,14 @@ component extends="coldbox.system.Interceptor" accessors="true"{
 		return this;
 	}
 
+    /**
+     * Create all RESTful routes for a resource
+     * @name The name of the resource.
+     * @handler The handler for the route. Defaults to the resource name.
+     * @parameterName The name of the id/parameter for the resource. Defaults to id.
+     * @only Limit routes created with only
+     * @except Exclude routes with except
+     */
     function addResource(
         required name,
         handler=arguments.name,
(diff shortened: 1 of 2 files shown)


lmajano:
WOW, this is in our roadmap, so awesome. Ok, I will need some time to review.

lmajano:
Thanks so much @murphydan I will be revising a few things based on this ticket: https://ortussolutions.atlassian.net/browse/COLDBOX-588

PR #313, opened 2017-05-09
Add only method to filter to only desired keys

elpete:
The `only` method lets you whitelist the keys you want returned from the request context.

Usage:

```
rc = { "email" = "john@example.com", "password" = "pass1234" };
var fields = event.only( [ "email", "username" ] );
// fields = { "email" = "john@example.com" };
```

### DIFF
--- a/system/web/context/RequestContext.cfc
+++ b/system/web/context/RequestContext.cfc
@@ -1215,15 +1215,20 @@ component serializable=false accessors="true"{
     	return ( getHTTPHeader( "X-Requested-With", "" ) eq "XMLHttpRequest" );
 	}
 
-	struct function only( required keys ) {
-		if ( isSimpleValue( keys ) ) {
-			arguments.keys = listToArray( keys );
+	/**
+	* Filters the collection or private collection down to only the provided keys.
+	* @keys A list or array of keys to bring back from the collection or private collection.
+	* @private Private or public, defaults public.
+	*/
+	struct function only( required keys, boolean private = false ){
+		if ( isSimpleValue( arguments.keys ) ){
+			arguments.keys = listToArray( arguments.keys );
 		}
 
 		var returnStruct = {};
-		for ( var key in keys ) {
-			if ( valueExists( key ) ) {
-				returnStruct[ key ] = getValue( key );
+		for ( var key in arguments.keys ){
+			if ( valueExists( key, private ) ){
+				returnStruct[ key ] = getValue( name = key, private = private );
 			}
 		}
 
--- a/system/web/context/RequestContext.cfc
+++ b/system/web/context/RequestContext.cfc
@@ -1215,6 +1215,21 @@ component serializable=false accessors="true"{
     	return ( getHTTPHeader( "X-Requested-With", "" ) eq "XMLHttpRequest" );
 	}
 
+	struct function only( required keys ) {
+		if ( isSimpleValue( keys ) ) {
+			arguments.keys = listToArray( keys );
+		}
+
+		var returnStruct = {};
+		for ( var key in keys ) {
+			if ( valueExists( key ) ) {
+				returnStruct[ key ] = getValue( key );
+			}
+		}
+
+		return returnStruct;
+	}
+
 	/************************************** RESTFUL *********************************************/
 
 	/**
--- a/tests/specs/web/context/requestcontextTest.cfc
+++ b/tests/specs/web/context/requestcontextTest.cfc
@@ -516,4 +516,15 @@
 			.toBe( { "name" = "John", "email" = "john@example.com" } );
 	}
 
+	function testOnlyPrivate() {
+		var event = getRequestContext();
+		event.setValue( "name", "John" );
+		event.setValue( "hackedField", "hacked!" );
+		event.setValue( "name", "Jane", true );
+		event.setValue( "hackedField", "hacked as well!", true );
+
+		expect( event.only( keys = "name,field-that-does-not-exist", private = true ) )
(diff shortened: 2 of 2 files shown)


bdw429s:
@elpete I love this Eric.  I saw something similar in a Laracast video and kind of wished ColdBox/ CF had this.  I was unclear though if the PHP quiv was part of Laravel or something baked into associative arrays in the PHP language.

What about also having `except()`?

What I'm not sure about is if this method should be on the `event` object.  It feels like it should be part of a struct, but CFML has no way to prototype that in.  Also, what would the private version of these be?

Perhaps we support this as a closure parameter to `getCollection()` and `getPrivateCollection()`?  I started to write out the code for that-- but it's really just about the same to use `struct.filter()` really...

elpete:
I see less of a need for this on the `prc` since I should be in control of all of those.  Also `except` seems like it could be a security concern (again, for the `rc`).  Those are just my quick thoughts.

I would love this on the `rc`, but unless you want to make it a [`CFCollection`](https://github.com/elpete/cfcollection), I'm not sure how to do it.

(Side note: I want CFML to have some sort of `ArrayAccess` interface l[ike PHP.](http://php.net/manual/en/class.arrayaccess.php))

system/web/context/RequestContext.cfc
```diff
@@ -1214,6 +1214,21 @@ component serializable=false accessors="true"{
     	return ( getHTTPHeader( "X-Requested-With", "" ) eq "XMLHttpRequest" );
 	}
 
+	struct function only( required keys ) {
+		if ( isSimpleValue( keys ) ) {
+			arguments.keys = listToArray( keys );
+		}
+
+		var returnStruct = {};
+		for ( var key in keys ) {
```
elpete: `structFilter` is CF10+.  Are we okay with that?

system/web/context/RequestContext.cfc
```diff
@@ -1214,6 +1214,21 @@ component serializable=false accessors="true"{
     	return ( getHTTPHeader( "X-Requested-With", "" ) eq "XMLHttpRequest" );
 	}
 
+	struct function only( required keys ) {
+		if ( isSimpleValue( keys ) ) {
```
elpete: I can add it, but I'm curious what is the benefit?

To me it's just noise.

bdw429s:
Putting the same comment here that I added to JIRA:

Should the method name be more specific so it's clear it applies to the request collection? That name would make more sense if the request collection was an object of its own, but the request context does many things so "only()" isn't very self explanatory that it is returning the collection. What about one of these two options:
```
event.getCollectionOnly( 'foo,bar,baz' );
event.getPrivateCollectionOnly( 'foo,bar,baz' );
```
or
```
event.getCollection( only='foo,bar,baz' );
event.getPrivateCollection( only='foo,bar,baz' );
```

elpete:
You are right that the right place for this would be `rc` if `rc` was an object.  I don't really have the same concerns with the method name.  In fact, the method names above seem long for me.  As for an additional argument to the `getCollection` method, that just seems like the wrong place for the method; hard to discover.  Maybe not, though.

bdw429s:
> I don't really have the same concerns with the method name. 
I guess I was just approaching it from this perspective.  Imagine you pick up the ColdBox framework and see the following event calls in a handler:
```
event.getCurrentHandler();
event.getHTTPHeader();
event.buildLink();
event.getCollection();
event.setView();
event.getHTMLBaseURL();
event.only();
```
Are you going to have any idea what that last line of code does?  It seems strangely abbreviated and without context.

jclausen:
I think Eric's point, that this would be best served as a "prototype" closure on the actual collection, is a valid one and is a more sound architecture pattern - making those closures specific to the collection it's filtering.  Having `only()` in the event object muddies the waters, a bit.

As a matter of fact, I'd love to see a subset of the cfCollection methods available as "prototype" closures on the `rc`/`prc`.

We would have to append that subset on the initial `rc`/`prc` creation, though, and it could create serialization headaches - not to mention the paradigm shift for some devs with using closures, in general.  It could also be inadvertently overwritten or removed.

It's a tough call, but if you want ColdBox to help push CFML forward, I think it's worthy of consideration.

elpete:
I see your point about the method being out of place.  I don't think it would be out of place at all if we could put it on the `rc` directly.  My second pick would be the `event.getCollection( only = [ "foo", "bar", "baz" ] )` — I just think it is a bit harder to discover and kind of muddies the original method.

bdw429s:
Yeah, I think we're all on the same page that having it on `rc` would be ideal, but I don't think that's in the realm of possibilities.  Swapping `rc` and `prc` out for a CFC instance instead of a struct would most certainly cause most issues than it solves.  I did something similar in my PropertyFile module, but it was a pain.

@lmajano Do you have input/care about this ? :)

lmajano:
Thinking ....

Luis Majano
CEO
Ortus Solutions, Corp
P/F: 1-888-557-8057

lmajano:
I would love for collections to have more umph! However, it would seem maybe appropriate for a module to accomplish that and intercept and add prototypes to the tc and prc. 

As for just filtering the collections I would probably defer then to the grtCollection() methods with the only arguments. 

With that said I would prefer the argument approach and also a separate module that can enhance cf collections with prototype like methods. The problem is that if you ineject method to a strict you loose the key. So maybe adding a _ approach would be nice.

bdw429s:
@lmajano Can a udf stored in a struct self reference the struct? I didn't think they have a "this" reference like JavaScript does.

elpete:
Yeah, I'd love an example of that, @lmajano.

PR #314, opened 2017-05-10
Fixed Adobe CF-11 compatibility bugs and also implemented intermediate variables to get around the ACF IsNull bug.

sanaullah:


system/interceptors/SES.cfc
```diff
@@ -124,6 +124,22 @@ component extends="coldbox.system.Interceptor" accessors="true"{
 			.setSESBaseURL( variables.baseURL );
 	}
 
+	// CF-11 include .cfm template can't access the methods which are only declare as property... (hack) have to create setter/getter methods
```
lmajano: @sanaullah the entire pull request looks good, except this portion of the setters.  Do you have a reproducible test case for this?

sanaullah:
Hi Luis,
my CF Version 11,0,12,302575 --- Update Level-12

I commented out the setters/getters methods to reproduce the error and attached screen shot of error detail
capture6
. 

At the moment current development branch is not compatible with CF-11, so I don't know how to reproduce the error before fixing other errors.

lmajano:
Wow I am so surprised this would be an issue. I am guessing we never saw
this because before we where in tags and no accessors.  So this seems like
an accessors bug.  Did you submit this to Adobe?

Totally defeats the purpose of implicit getters and setters, what a drag!

Luis F. Majano
CEO
Ortus Solutions, Corp
www.ortussolutions.com

ColdBox Platform: http://www.coldbox.org
Linked In: http://www.linkedin.com/pub/3/731/483
Blog: http://www.luismajano.com
IECFUG Manager: http://www.iecfug.com
Social: twitter.com/lmajano facebook.com/lmajano

bdw429s:
So, to clarify, the accessor method generated on the component aren't available in the variables scope of an included template?  Def deserves a bug report for Adobe. if there's not one.

bdw429s:
I entered a ticket here complete with a test repo that shows the issue.  It is still present in ACF 2016
https://tracker.adobe.com/#/view/CF-4198705

PR #315, opened 2017-05-17
Add new request context dsl injection.

elpete:
To get the request-scoped instance, simply use the provider: `coldbox:requestContext`.

If you are using this in anything other than request-scoped or other transient-scoped, use a provider: `provider:coldbox:requestContext`.  A reminder that methods will be forwarded on to the provided instance.  That means:

```cfc
component {
    property name="event" inject="provider:coldbox:requestContext";

    function doStuff() {
        var rc = event.getCollection();
        // same as
        var rc = event.get().getCollection();
    }
}
```

works just fine, with or without calling `get` first.

### DIFF
--- a/system/ioc/dsl/ColdBoxDSL.cfc
+++ b/system/ioc/dsl/ColdBoxDSL.cfc
@@ -102,17 +102,6 @@ Description :
 							// just get setting
 							return instance.coldbox.getSetting( thisLocationKey );
 						}
-						case "requestContext"       : {
-							if ( thisLocationKey == "rc" ) {
-								return instance.coldbox.getRequestService().getContext().getCollection();
-							}
-							else if ( thisLocationKey == "prc" ) {
-								return instance.coldbox.getRequestService().getContext().getPrivateCollection();
-							}
-							else {
-								instance.log.debug( "The collection type requested: #thisLocationKey# does not exist in the request context. Valid collection types are [rc, prc]." );
-							}
-					    }
 						case "modulesettings"		: {
 							moduleSettings = instance.coldbox.getSetting("modules");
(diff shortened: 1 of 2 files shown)


bdw429s:
@elpete @lmajano Are we going to do the change to automatically wrap this in a provider?

lmajano:
I am not sure it is worth it.

bdw429s:
@lmajano I don't believe there was any great deal of complexity to implementing it if that's what you're concerned about.  I'm just afraid that we're going to make it really easy for people to screw up with this injection of a request-scoped item.  

I'm also a little confused since I thought the consensus on Slack was that we would implement the automatic provider, but then you merged this pull without that feature without any further discussion.

elpete:
I had the opposite recollection — that we would leave the automatic provider alone. That way no one would provider a provider.

lmajano:
Yes.

I was just doing what eric said :)

Luis Majano
CEO
Ortus Solutions, Corp
P/F: 1-888-557-8057

bdw429s:
Ok, weird.  I was going off this conversation, which I thought was framed with the idea that it was going to be a part of this pull.

```
Brad Wood
[1:50 PM] 
@lmajano Are you ok with moving forward with the idea to enhance the inbuilt WireBox provider to now accept a UDF (as mutex with name and DSL) where that UDF (closure) is executed to produce the instance being provided?
This will allow us to wrap providers around instances that need a bit of building on the fly first.
... discussion ...
Luis Majano [4:54 PM] 
ohh the actual Provider.cfc
ok now I get it
I am ok with that
now I get it
...
Eric Peterson [6:26 PM] 
Ha. Give me a ticket or else I'll forget.
```

PR #317, opened 2017-05-30
Fix 597

Tropicalista:


lmajano:
Which ticket is this for @Tropicalista

lmajano:
Forget it found it

system/core/dynamic/HTMLHelper.cfc
```diff
@@ -86,7 +86,7 @@ Description :
 				if( NOT listFindNoCase(event.getValue(name="cbox_assets",private=true),thisAsset) ){
 
 					// Load Asset
-					if( findNoCase(".js", thisAsset) ){
+					if( findNoCase( "js", listLast(thisAsset,".") ) ){
```
bdw429s: This won't work if a value in the query string has a period in it.
```site.com/assets/myFile.js?foo=bar.baz```
See my comments in the Box team Slack channel.
cc/ @lmajano

lmajano:
So I see two pull requests with the same ticket? Which one is the one to use?

PR #318, opened 2017-06-01
Reset the response when calling setup

elpete:
In order to prevent things like duplicate headers, reset the response when calling setup in between tests.


elpete:
This call could make more sense in the `removeContext` method.  Thoughts?

lmajano:
I think this is in the right location

PR #320, opened 2017-06-03
Fix 597

Tropicalista:
Add an extension check to prevent that css path containing .js string is added as js

Ex https://cdnjs.cloudflare.com/ajax/libs/gridstack.js/0.3.0/gridstack-extra.min.css

### DIFF
--- a/system/core/dynamic/HTMLHelper.cfc
+++ b/system/core/dynamic/HTMLHelper.cfc
@@ -1,4 +1,4 @@
-﻿<!-----------------------------------------------------------------------
+<!-----------------------------------------------------------------------
 ********************************************************************************
 Copyright Since 2005 ColdBox Framework by Luis Majano and Ortus Solutions, Corp
 www.ortussolutions.com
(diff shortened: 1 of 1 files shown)


bdw429s:
Hi @Tropicalista , did you test this latest version?  I don't see how it would work.  If you didn't pass the fileExt parameter, `isJS` would default to false (or if you explicitly passed `css` the same would be true), but the later if statement would still treat it as JS anyway because the string `.js` still appears in the string and `findNoCase(".js", thisAsset)`  will evaluate as true.  

I would expect that if there is a passed file type, it completely overrides any "guessing".  

And finally, are you going to do the change to correct check the end of the URL for an extension.  Something along the lines of:
```js
listLast( listFirst( listFirst( thisAsset, '##' ), '?' ), '.' )
```
Here's a cleaner but more verbose one that uses some Java libs for URL parsing.
```java
var theURL = createObject( 'java', 'java.net.URL' ).init( thisAsset );
var fileUtils = createObject( 'java', 'org.apache.commons.io.FilenameUtils' );
fileUtils.getExtension( theURL.getPath() );
```

lmajano:
So I see two pull requests with the same ticket? Which one is the one to use?

Tropicalista:
No one is completed, I'd try to make a pull request following brad's suggestions.

lmajano:
Should I close this, or we revamping this?

Tropicalista:
Please close this. I will send another pr.

PR #323, opened 2017-07-27
Ensure modules are only activated once, even if it is a child module of one of its dependencies.

elpete:
Fixes [COLDBOX-604](https://ortussolutions.atlassian.net/browse/COLDBOX-604)


jclausen:
Good catch, @elpete!

PR #325, opened 2017-08-09
Fixed: overrideEvent() not working from proxy context

sethstone:
This appears to have been a regression when converting Controller.cfc to cfscript format in this commit: e969c3742ca34b6ad8a05bfc946caec85a01fbc7

### DIFF
--- a/system/remote/ColdboxProxy.cfc
+++ b/system/remote/ColdboxProxy.cfc
@@ -121,7 +121,7 @@ Description :
 
 			//Execute the Event if not demarcated to not execute
 			if( NOT event.isNoExecution() ){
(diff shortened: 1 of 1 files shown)


bdw429s:
Good catch, it looks like we renamed the argument from `default` to `defaultEvent` to avoid syntax errors in Adobe CF, but failed to fix every place calling the function to correct their argument names.

bdw429s:
@sethstone Can you please put in a ticket into JIRA for this fix?

sethstone:
@bdw429s Here's the ticket: https://ortussolutions.atlassian.net/browse/COLDBOX-608  (Let me know if I need to take it further through the process)

PR #327, opened 2017-09-14
Update FileAppender.cfc

peterp01:
Add extraInfo to the message before clean-up. This allows log files to be imported correctly as csv if there are " in the extraInfo (like with json)


lmajano:
Thanks mi amigo!!!

PR #331, opened 2017-11-24
Adds getPageContextResponse() helper method to deal with ACF2016

jclausen:
Adds a helper method to handle ACF's overload of the page context response.


lmajano:
merged manually! Thanks

PR #332, opened 2017-12-01
Coldbox 639 and 640 HtmlHelper labelAttrs and inputInsideLabel

MordantWastrel:


lmajano:
Thanks @MordantWastrel I will start reviewing this.

PR #335, opened 2017-12-09
Add check for eCacheEntry.provider key

jclausen:
Ensures support for backward compatibility until `HandlerService.cfc:209` and `RequestService.cfc:139` are updated to set the provider.


lmajano:
merged manually

PR #337, opened 2017-12-11
Move system instantiating inline to work with mixins

elpete:


elpete:
To test it with a mock.

bdw429s:
@elpete @lmajano Would it be better to simply provide a `getSystem()` method and then mock that method in our tests.  Adding it to the method signature doesn't seem right-- at least not just to allow mocking.  It also doesn't make much sense to re-create `java.lang.System` every time you need it since it only allows static access and is never instantiated.

lmajano:
@elpete thoughts?

PR #338, opened 2017-12-12
Optimizations for grabbing java System in Util

elpete:
See [this comment](https://github.com/ColdBox/coldbox-platform/pull/337#issuecomment-350889290).


lmajano:
This might break the injections though that are used for module config and ColdBox config since you now need to inject the other method.

I have a funky idea.  Why not declare these methods into the `this` scope as closures so they can bring in their environment with them.  Making them like pseudo static assets?

PR #340, opened 2018-01-10
Add subdomain routing capability

elpete:
A domain can be specified on a route or with closure.
The incoming domain must match the domain in order for the route to be a match.
Additionally, a wildcard or parameter can exist in the route, like `:username.forgebox.io`.
The value of `username` (in this case) will exist in the `rc` scope.

Examples:

```
// will verify the domain matches before matching
addRoute(
    domain = "subdomain-routing.dev",
    pattern = "/",
    handler = "subdomain",
    action = "index"
);

// adds a username route parameter that is available in the `rc` scope when matched
with( domain = ":username.forgebox.io" )
    .addRoute( pattern = "/", handler = "dashboard", action = "index" )
.endWith();
```

> **Note:** this PR will fail until https://github.com/Ortus-Solutions/TestBox/pull/63 is merged.

### DIFF
--- a/system/interceptors/SES.cfc
+++ b/system/interceptors/SES.cfc
@@ -5,7 +5,7 @@
 * This interceptor provides complete SES and URL mappings support to ColdBox Applications
 */
 component extends="coldbox.system.Interceptor" accessors="true"{
-	
+
 	/**
 	 * The routing table
 	 */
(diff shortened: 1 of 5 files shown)


elpete:
@lmajano What is the status on this?

lmajano:
In process

elpete:
Note that even though the TestBox change was merged, unless this project is being testing against the latest bleeding edge release of TestBox, these tests will fail.

lmajano:
Merged manually! gracias!

PR #345, opened 2018-02-01
Convert ioc injector

Yamaha32088:


Yamaha32088:
I removed the `cfinvoke` functionality from the functions `processAfterCompleteDI` and `injectTarget`. I am calling the functions directly now, it seemed to be the easiest way to get rid of the Lucee 4.5 error. Please let me know if you don't like this way and I can find an alternative.

Yamaha32088:
It appears the build hung on Lucee 4.5. I don't see an option where I can restart the build.

lmajano:
Thanks so much.merged manually as I found other issues.  But looks great!
https://github.com/ColdBox/coldbox-platform/commit/26824d1e764f4fb22dd9531333626dc23e891e7b

PR #347, opened 2018-02-16
Added Local InvalidHTTPMethod call for SES Routes

gpickin:
Allowed methods in the handler was supported and was throwing the handlers onInvalidHTTPMethod call, but if the allowed method was not supported by the routes, this handlers onInvalidHTTPMethod was not being called.

CC @elpete


lmajano:
@gpickin can I have a test on this, I have no way to make sure this works. thanks

lmajano:
@gpickin can I have a test on this, I have no way to make sure this works. thanks if not I can't add it.

lmajano:
closing this, if you need it reopen let me know, I think this is ok now.

PR #357, opened 2018-04-19
Fix a requirement for columnName if column is defined

gpickin:
Without this, if you set a column, but not columnName, it will not use the default 'name' field, it will overwrite it with the column field ( value ) because of the else.
Tested locally, this works.

### DIFF
--- a/system/core/dynamic/HTMLHelper.cfc
+++ b/system/core/dynamic/HTMLHelper.cfc
@@ -1118,7 +1118,7 @@ Description :
 				// struct normalizing
 				if( isStruct( val[x] ) ){
 					// Default
-					thisName = thisValue;
+					thisName = "";
 
 					// check for value?
 					if( structKeyExists(val[x], "value") ){ thisValue = val[x].value; }
@@ -1131,8 +1131,10 @@ Description :
 					if( len( arguments.nameColumn ) ){
 						if( structKeyExists( val[x], arguments.nameColumn ) ){ thisName = val[x][nameColumn]; }
 					}
-					else{
-						if( structKeyExists( val[x], arguments.column ) ){ thisName = val[x][column]; }
+					
+					// Only set to the value if thisName is still empty
+					if( thisName == "" ){
+						thisName = thisVal;
 					}
 
 				}


system/core/dynamic/HTMLHelper.cfc
```diff
@@ -1131,8 +1131,10 @@ Description :
 					if( len( arguments.nameColumn ) ){
 						if( structKeyExists( val[x], arguments.nameColumn ) ){ thisName = val[x][nameColumn]; }
 					}
-					else{
-						if( structKeyExists( val[x], arguments.column ) ){ thisName = val[x][column]; }
+					
+					// Only set to the value if thisName is still empty
```
lmajano: What about the if statement that checks if the value has a `column` field? You ignore it

gpickin: No, that still does it’s thing.

lmajano: Where? The code is removed

lmajano:
Yes, but the else is also necessary if not it won't discover it via the column code:
`if( structKeyExists( val[x], arguments.column ) ){ thisName = val[x][column]; }`

aarongreenlee:
@gpickin thanks for your pull request. Would you like to finalize it by making the adjustment @lmajano proposed?

Respectfully,

-A

gpickin:
The else was a bug.
But i wonder if we should use the `thisName = thisValue` instead of `thisVal` following the else that I removed.
Leaving the ELSE in there, it would mess things up. 
The ELSE meant that if `arguments.nameColumn` did not have a length, it would overwrite `theName` when that is not what we want. If theName was set elsewhere, we want to preserve it, and only use `thisValue` if none of the other issues 

If you view the full file ( not just the preview ) you will see the full function, and it might be clearer.

Now a conflict has arisen too.

gpickin:
It was moved into the HTMLHelper module.

gpickin:
I'll fix it in the new location, and we can close this one.

My points are still valid. I came across this when data was missing when it shouldn't be... so its a bug.

gpickin:
Check out the new pull
https://github.com/ColdBox/coldbox-platform/pull/370

PR #360, opened 2018-04-24
Fix virtual inheritence to inject in to both the variables and the `this` scope

elpete:


lmajano:
Merged manually due to some code updates
https://ortussolutions.atlassian.net/browse/WIREBOX-76

PR #363, opened 2018-06-30
Resolves COLDBOX-708

jclausen:
prevents rc/prc explicit values from being overwritten by paramed route rc/prc values. Also adds keys for persisting the `overwrite` argument to the Router `rc()` and `prc()` methods.

### DIFF
--- a/system/web/routing/Router.cfc
+++ b/system/web/routing/Router.cfc
@@ -1032,7 +1032,9 @@ component accessors="true" extends="coldbox.system.FrameworkSupertype" threadsaf
 			verbs 					= "",
 			headers 				= {},
 			rc 						= {},
-			prc 					= {}
+			prc 					= {},
+			overwriteRC             = {},
+			overwritePRC            = {}
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


lmajano:
I updated it to not override, it should have not overriden it

jclausen:
That is the expected behavior.  My pull request was updated to address the fact that the `overwrite=true` argument passed to the `rc()` and `prc()` methods does not persist beyond the initial route configuration and would really only be applicable if two of the same params are passed in to a route configuration  - which doesn't seem to have any value.

It seem to me that passing a flag of `overwrite=true` to those methods should signify that the value I am passing is authoritative when the rc/prc values are appended. If you look at my commit history on this PR, you'll see that I started out just setting the overwrite on `append()` to `false` and then updated it to persist that setting.

One other thing I think should be considered for the future, is the ability to pass a closure in as the value argument to `rc` and `prc`.  I could see this being used to sanitize or convert inbound values - like SES slugs:

```
.rc( 
   "id", 
   function( id ) {
     if( !isValid( "GUID", id ) ){
         return myModel.findBySlug( id ).getId();
     }
   }
);
```

PR #364, opened 2018-07-03
Fix COLDBOX-711

Tropicalista:
Let user choose what to encode otherwise he cannot add fontawesome icons.

### DIFF
--- a/system/core/dynamic/HTMLHelper.cfc
+++ b/system/core/dynamic/HTMLHelper.cfc
@@ -1860,7 +1860,7 @@ component extends="coldbox.system.FrameworkSupertype" singleton{
 			arguments,
 			"value,label,wrapper,labelWrapper,groupWrapper,labelClass,inputInsideLabel",
 			buffer
-		).append( ">#encodeForHTML( arguments.value )#</button>" );
+		).append( ">#arguments.value#</button>" );
 
 		//wrapper?
 		wrapTag( buffer, arguments.wrapper, 1 );


lmajano:
Can you expand on why you can’t use font awesome in the value?

bdw429s:
We added the encoding in the latest version for security by default, however I always had suggested that the HTML helper should provide two methods-- one that encodes and one that doesn't.  Just like jQuery does for instance when appending HTML or plain text.

lmajano:
@bdw429s can you expand no how that would look like?

Tropicalista:
You could add another parameter "htmlValue". Also note that this affect also ContentBox form, just give a look at SEO tab on page editor

MordantWastrel:
"All HTML output is now XSS encoded for attributes and tag content."

isn't this our job? Why is Coldbox doing this for us? it'd be swell to have it be an option because I can think of plenty of use cases, but unfortunately we also have plenty of use cases where HTMLHelper is now broken because things are encoded that didn't need to be encoded (and that aren't exposed to any request variables).

lmajano:
I think it is ok to revert this change to OFF by default and unless you specifically turn it on, then turn it on.

Tropicalista:
Why not to put the cncoder on helper function and add "encoded" parameter to evaluate if should encode or not? seems a good solution to me

lmajano:
@Tropicalista can you expand on your solution?

Tropicalista:
Something like:

```
function getString(myString,shouldEncode){
    if(shouldEncode){
        return encodeForHTML(myString)
    }
    return mystring;
}
```

then in every helper function

`
	function heading( required content, numeric size=1, shouldEncode=false ){
		return "<h#arguments.size#>#getString( arguments.content, arguments.shouldEncode )#</h#arguments.size#>";
	}
`

lmajano:
https://ortussolutions.atlassian.net/browse/COLDBOX-714

I have reverted this to NOT encode by default.  YOu can now use module settings to fine tune the HTML Helper:

```
moduleSettings = {

  htmlHelper = {
    encodeValues = true
  } 

}
```

PR #370, opened 2018-09-09
Fix a requirement for columnName if column is defined - new

gpickin:
Without this, if you set a column, but not columnName, it will not use the default 'name' field, it will overwrite it with the column field ( value ) because of the else.
Tested locally, this works.

Update of pull request after new file location and conversion of tag to script.
https://github.com/ColdBox/coldbox-platform/pull/357

### DIFF
--- a/system/modules/HTMLHelper/models/HTMLHelper.cfc
+++ b/system/modules/HTMLHelper/models/HTMLHelper.cfc
@@ -1656,8 +1656,7 @@ component extends="coldbox.system.FrameworkSupertype" accessors=true singleton{
 
 			// struct normalizing
 			if( isStruct( val[ x ] ) ){
-				// Default
-				thisName = thisValue;
+				thisName = "";
 
 				// check for value?
 				if( structKeyExists(val[ x ], "value") ){ thisValue = val[ x ].value; }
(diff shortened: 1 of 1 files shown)


system/modules/HTMLHelper/models/HTMLHelper.cfc
```diff
@@ -1670,8 +1669,10 @@ component extends="coldbox.system.FrameworkSupertype" accessors=true singleton{
 				if( len( arguments.nameColumn ) ){
 					if( structKeyExists( val[ x ], arguments.nameColumn ) ){ thisName = val[ x ][nameColumn]; }
 				}
-				else{
-					if( structKeyExists( val[ x ], arguments.column ) ){ thisName = val[ x ][column]; }
+				
+				// If thisName is still the default, use the content of thisValue as the name
```
lmajano: What about this statement:
`if( structKeyExists( val[ x ], arguments.column ) ){ thisName = val[ x ][column]; }`

This checks if the column exists in the struct, if it does is uses it as the default value.

gpickin: Read the other pull request.
This is not what we want.

This would override other possible names/values we should use instead.
This is a BUG, this else is a bug

PR #371, opened 2018-09-09
Fix ColdBox-597

Tropicalista:
Prevent wrong filetype attribution when an url contains ".js"

### DIFF
--- a/system/modules/HTMLHelper/models/HTMLHelper.cfc
+++ b/system/modules/HTMLHelper/models/HTMLHelper.cfc
@@ -116,7 +116,7 @@ component extends="coldbox.system.FrameworkSupertype" accessors=true singleton{
 			} )
 			.each( function( item ){
 				// Load Asset
-				if( findNoCase( ".js", item ) ){
+				if( listLast( listFirst( listFirst( item, '##' ), '?' ), '.' ) EQ 'js' ){
 					sb.append(
 						'<script src="#jsPath##encodeForHTMLAttribute( item )#" #asyncStr##deferStr#></script>'
 					);


system/modules/HTMLHelper/models/HTMLHelper.cfc
```diff
@@ -116,7 +116,7 @@ component extends="coldbox.system.FrameworkSupertype" accessors=true singleton{
 			} )
 			.each( function( item ){
 				// Load Asset
-				if( findNoCase( ".js", item ) ){
+				if( listLast( listFirst( listFirst( item, '##' ), '?' ), '.' ) EQ 'js' ){
```
lmajano: I fail to see how the `findNoCase()` wouldn't still find a `.js` in the path since it is an open search of the string.  I am not sure I follow the pull request, can you expand

Tropicalista: HI @lmajano can you try this gist:

    item = 'cdnsource/gridstack.js/style.min.css';
    isJs = findNoCase( ".js", item );
    isJs2 =  listLast( listFirst( listFirst( item, '##' ), '?' ), '.' ) EQ 'js';
    writedump(isJs);
    writedump(isJs2);

Tropicalista:
Luis the problem here is not the search of the string, if the correct file attribution. Please can you check the gist I posted?

If you try the code with

`
https://cdnjs.cloudflare.com/ajax/libs/gridstack.js/0.4.0/gridstack-extra.min.css
`

the old code will add this to js, but it is actually a css.

system/modules/HTMLHelper/models/HTMLHelper.cfc
lmajano: I totally get it now.  Tricky.  TRhanks

PR #373, opened 2018-09-10
Allow closure for COLDBOX_FAIL_FAST

sjdaniels:
No idea why there is that strange commit from 2014 in here. Sorry about that. I tried to rebase my fork but doesn't appear to get rid of it.


lmajano:
Merged manually!!! Thanks!1

PR #375, opened 2018-09-12
Use setLength(0) since clear is not a method on StringBuilder

elpete:


bdw429s:
@elpete Is there a ticket for this?  If not, can you create one pls?  The only failing engine is 2018 which has some major issues still unfixed.

PR #382, opened 2018-11-08
COLDBOX-744 Use a simple array for 'event pool' to avoid locking issu…

DominicWatson:
…es with Syncrhonized maps. This DOES lead to a slightly slower performance when wanting to GET specific interceptos from the map; HOWEVER, this is not a common operation (the most common is just to get the entire pool and loop over it, which carries on working fine with an array).

### DIFF
--- a/system/web/context/InterceptorState.cfc
+++ b/system/web/context/InterceptorState.cfc
@@ -4,7 +4,7 @@
 * ---
 * This object models an interception state
 */
-component accessors="true" extends="coldbox.system.core.events.EventPool"{
+component accessors="true" {
 
 
 	/**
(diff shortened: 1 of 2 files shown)


lmajano:
Thanks reviewing.

lmajano:
@DominicWatson I have removed the syncrhonized piece from the original source.  I prefer to stick with a hash map as lookups are way faster than iterating and finding states in an array.  I don't see much point on syncronized the map since mostly we do reads, so I have committed it, try it out, and if it satisfies the ticket, we can go with that.

lmajano:
https://github.com/ColdBox/coldbox-platform/commit/d7a8f445d807edd7c32ac9ed8f7f53cdd8f9b3c2

DominicWatson:
Sure thing, good to close from me +1

PR #383, opened 2018-11-09
Throw a nicer DSL error if ColdBox is not linked

gpickin:


bdw429s:
@gpickin Can you create a ticket for this change and link here?

PR #385, opened 2018-11-15
COLDBOX-746 multiple handler, view and layout locations

DominicWatson:
This pull request alters the core to allow for multiple external handler, layout and view locations. In addition, when multiple handler locations are used, each location can provide different actions and all unique combinations of handler/action will then be accessible.


lmajano:
@DominicWatson This will take me also time since it changes tons of core pieces.  Also, is there a way you can add more tests to provide more coverage for your changes.  If not, I have no way of testing the features.

lmajano:
Sorry, @DominicWatson this has taken me very long to get to. Still working on it.

lmajano:
@DominicWatson I think I totally missed this on the run to do ColdBox 6.  If this is something still of interest, can you rebase and re send please?

PR #386, opened 2018-11-15
COLDBOX-748: use dummy flash scope when sessions disabled

DominicWatson:
This is really just a safety backstop. We sometimes disable sessions for things like BOT requests and this logic defends against odd code that might try accessing the flash scope during these requests.

### DIFF
--- a/system/web/services/RequestService.cfc
+++ b/system/web/services/RequestService.cfc
@@ -211,7 +211,13 @@ component extends="coldbox.system.web.services.BaseService"{
 	 * Return the flash scope instance in use by the framework.
 	 */
 	any function getFlashScope(){
-		return variables.flashScope;
+		var appSettings = getApplicationMetadata();
+
+		if ( IsBoolean( appSettings.sessionManagement ?: "" ) && appSettings.sessionManagement ) {
+			return variables.flashScope;
+		}
+
+		return variables.mockFlashScope;
 	}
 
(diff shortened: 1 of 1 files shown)


jclausen:
I totally support a fallback when sessions are disabled. Is there any reason we couldn't point this to the `ColdboxCacheFlash` instead, so that it can persist between requests?  There are already checks in the cache flash for whether or not sessions are enabled.

lmajano:
Yep I think it’s best to have the cache as callback.

lmajano:
Fallback that is

DominicWatson:
Sounds good to me - would you like me to adjust the PR?

jclausen:
If you could, Dom, that would be great!  Should be just changing two lines.

lmajano:
Actually I am wondering why in the getter and not in the constructor of the flash. It will be unperformant to leave at the getter

DominicWatson:
Thinking on this more, if an application will dynamically turn on/off session management per request - it could just use the Cache storage instead of session storage...

You already have logic here for checking session management when initializing this so I think I should just close this PR unless you think otherwise?

lmajano:
I agree @DominicWatson , I am not sure if we need this.

DominicWatson:
Closing, thanks for pointing out the CacheFlash :man_facepalming:

PR #389, opened 2019-01-11
Add test and fix for adding a LogBox category after the fact

elpete:
Adding a category after LogBox had already been configured errored because the `*` appenders wasn't expanded.  This PR expands the appenders in the `category` method, if needed.


lmajano:
Gracias @elpete

PR #393, opened 2019-01-25
Add 2018 Beta 2 Back to testing matrix

jclausen:


jclausen:
This will not be able to be merged until the WAR file for the Beta is updated with the latest MySQL driver.  In addition, `createWebService` calls currently fail due to the usage of the class `com.sun.tools.javac.Main`

PR #397, opened 2019-02-28
Add Elixir version 3 path method.

elpete:
Uses a version flag that defaults to 3 to avoid adding a new method.


lmajano:
I merged it too soon, it doesn't work and no tests where added, I am adding the tests and the key doesn't work.

lmajano:
`var key = right( href, len( href ) - 1 );`
Is basically only removing the first character of the href.

PR #398, opened 2019-03-02
Map : to / in URL

richardherbert:
When an event includes a module the separator is ":" and should be mapped to "/" in the translated URL. There could be other places in the codebase where this change should also be applied but I don't have enough knowledge to make that decision :-)

### DIFF
--- a/system/web/Controller.cfc
+++ b/system/web/Controller.cfc
@@ -345,7 +345,7 @@ component serializable="false" accessors="true"{
 			// Default event relocations
 			case "SES" : {
 				// Route String start by converting event syntax to / syntax
-				routeString = replace( arguments.event, ".", "/", "all" );
+				routeString = replaceList( arguments.event, ".,:", "/,/" );
 				// Convert Query String to convention name value-pairs
 				if( len( trim( arguments.queryString ) ) ){
 					// If the routestring ends with '/' we do not want to
--- a/system/web/Controller.cfc
+++ b/system/web/Controller.cfc
@@ -344,8 +344,16 @@ component serializable="false" accessors="true"{
 
 			// Default event relocations
 			case "SES" : {
+				// Convert module into proper entry point
+				if( listLen( arguments.event, ":" ) > 1 ) {
+					var mConfig = getSetting( "modules" );
+					var module = listFirst( arguments.event, ":" );
+					if( structKeyExists( mConfig, module ) ){
+						arguments.event = mConfig[ module ].inheritedEntryPoint & "/" & listRest( arguments.event, ":" );
+					}
+				}
 				// Route String start by converting event syntax to / syntax
-				routeString = replaceList( arguments.event, ".,:", "/,/" );
+				routeString = replace( arguments.event, ".", "/", "all" );
 				// Convert Query String to convention name value-pairs
 				if( len( trim( arguments.queryString ) ) ){
 					// If the routestring ends with '/' we do not want to
@@ -716,10 +724,7 @@ component serializable="false" accessors="true"{
 				}
 				// Around Handler Advice Check?
 				else if(
-					!arguments.prePostExempt
-					&&
-					oHandler._actionExists( "aroundHandler" )
-					&&
+					oHandler._actionExists( "aroundHandler" ) AND
 					validateAction( results.ehBean.getMethod(), oHandler.aroundHandler_only, oHandler.aroundHandler_except )
 				){
 					results.data = oHandler.aroundHandler(
--- a/system/web/context/RequestContext.cfc
+++ b/system/web/context/RequestContext.cfc
@@ -1078,7 +1078,15 @@ component serializable=false accessors="true"{
 
 			// Translate link or plain
 			if( arguments.translate ){
-				arguments.to = replaceList( arguments.to, ".,:", "/,/" );
+				// Convert module into proper entry point
+				if( listLen( arguments.to, ":" ) > 1 ) {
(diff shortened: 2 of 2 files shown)


richardherbert:
Like we discussed on Slack, I appreciate that the issue is more complex than I can comprehend but I think the inconsistent way in which a module event is translated needs addressing. Unfortunately I don't have the framework knowledge to take this any further than I have. Maybe I'll have to be content with creating my own Request Context/Controller Decorators as you suggested on Slack.

bdw429s:
Yes, of course it needs addresses but let's do it right. I think the final solution is much easier than you think it is.  Look, this is all you need to turn a module name into the entry point, taking into account any nested modules:

```js
getModuleConfig( 'moduleName' ).inheritedEntryPoint
```

That means the entire solution is probably something as simple as this:
```js
// Convert module into proper entry point
if( listLen( arguments.event, ':' ) > 1 ) {
  arguments.event = getModuleConfig( listFirst( arguments.event, ':' ) ).inheritedEntryPoint & '/' & listRest( arguments.event, ':' );
}
```

richardherbert:
I'm well out of my depth here but I'll have a go.

So I've added your suggested code to both `Controller.cfc` and `RequestContext.cfc` but in both cases the method `getModuleConfig()` is not found. How should I inject it into these components?

bdw429s:
The `getModuleConfig()` method comes from the framework super type which I guess isn't there for the controller or request context.  It's easy to reproduce though.  Here is what that method does which is just getting a struct from the ColdBox settings.

https://github.com/ColdBox/coldbox-platform/blob/development/system/FrameworkSupertype.cfc#L300-L308

richardherbert:
Okay, that seems to work. Is it alright to dup that code into both of those components rather than some fancy reuse injection?

If so, do you have some convention on where I should poke it?

bdw429s:
Unless Luis says otherwise, I would just copy the `getSetting()` line and the structkeyexists lines in both places.  If the struct key doesn't exist, the logic can probably just ignore the module name.

richardherbert:
Okay, let me take a look at that.

bdw429s:
@richardherbert Scanning through, this look pretty good.  Now all we need are a couple tests to cover the new scenario.  Are you interested in trying to create them?  Here are the existing tests for build link:

https://github.com/ColdBox/coldbox-platform/blob/development/tests/specs/web/context/requestcontextTest.cfc#L387-L441

lmajano:
Sorry, just started reading this. 

I see where getting the entry point can help, but what if you are leveraging relocations by convention, then you will get a result that is not the same.  So the question is, when you use relocate, should you be smart enough to do a lookup for the module, or take it by convention? If you are using the `module:` prefix, most likely this is a convention route and not a direct entry point route.

Also, if you want smart lookups you can use named routes instead.  So in all reality, I am not sure if we should do the lookups because it could completely change the way the routing will be created in expectance of the user's input.

Thoguhts

bdw429s:
@lmajano Sorry, I really don't understand anything you said in your comment.  Build link and relocate can both accept an event in the internal event syntax and when `translate` is true, those functions should correctly translate the event to the route.  I don't think there would be cases where a colon would otherwise be in the URL.  Can you provide some existing scenarios that you think this code would adversely affect.

lmajano:
What I mean, is that the buildLink() and relocate() take the event string, which is expected to just be outputted in ses format.  If we change the module name to an entry point, it can cause issues, because the entry point can be different than the module name.

`/module/handler/action` is by convention, which can be different than `/entrypoint/handler/action`.

So what I am saying, is that if we do lookups, the expectation of what you write can be different than the output.  So I am concerned with doing lookups for that reason.

bdw429s:
@lmajano But you seem to be forgetting, this method actually has a `translate` parameter, whose job is to take the internal event syntax and turn it into the default SES convention route.  So the precedent (and feature) is already there to convert non-SES to SES.  We are not changing that, just completing it so it actually works correctly with modules.  

I honestly just don't understand what your concern is.  You pointed out that the module name can be different from the entry point.  Yes! _That's the whole reason we're translating it so it will work!_  There is no value in the current behavior of the `translate` flag which doesn't handle modules at all.  Let's make translate work, and work correctly.  Swapping the name of the module with the proper entry point.

Again, can you provide some existing scenarios that you think this code would adversely affect? I can't think of any.

bdw429s:
@lmajano Can I merge this?

lmajano:
Still revising

richardherbert:
I've just started to write some test coverage and this has raised an issue.

I'm getting an error "_The setting modules does not exist_" which suggests that https://github.com/ColdBox/coldbox-platform/blob/02a55af0a9d00ee95c9f47f90b65f30e0e2aedb8/system/web/Controller.cfc#L349 is failing due to the Coldbox.cfc used by the test harness not having a `modules` section or that there is no default `modules` section in the ColdBox framework.

I could be wrong.

Do you want me to commit/push the failing tests I have?

bdw429s:
@richardherbert If you're modifying the test above that I linked to, it extends `BaseModelTest` so it doesn't have ColdBox loaded.  You'll need to mock the `getSetting()` call to return what you want for the sake of the test.

Here's an example of a test that fakes a method:
https://github.com/ColdBox/coldbox-platform/blob/development/tests/specs/web/context/requestcontextTest.cfc#L90

lmajano:
I merged all your request, but had to do it manually for some reason @richardherbert I apologize for that.  Ticket is here https://ortussolutions.atlassian.net/browse/COLDBOX-791

PR #401, opened 2019-03-24
Update Controller.cfc

angel-chrystian:
Skip fileExists check if the incoming path has a ":", because it causes error in sandbox in Windows.

### DIFF
--- a/system/web/Controller.cfc
+++ b/system/web/Controller.cfc
@@ -798,11 +798,11 @@ component serializable="false" accessors="true"{
 		var foundPath 	= "";
 
 		//Check 1: Inside of App Root
-		if ( fileExists( variables.appRootPath & arguments.pathToCheck) ){
+		if ( !find( ':', arguments.pathToCheck) and fileExists( variables.appRootPath & arguments.pathToCheck) ){
 			foundPath = variables.appRootPath & arguments.pathToCheck;
 		}
 		//Check 2: Expand the Path
-		else if( fileExists( expandPath( arguments.pathToCheck ) ) ){
+		else if( !find( ':', arguments.pathToCheck) and fileExists( expandPath( arguments.pathToCheck ) ) ){
 			foundPath = expandPath( arguments.pathToCheck );
 		}
 		//Check 3: Absolute Path
--- a/system/web/services/RoutingService.cfc
+++ b/system/web/services/RoutingService.cfc
@@ -743,7 +743,7 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
 				thisFolder = listgetAt(rString,x,"/" );
 
 				// Check if package exists in convention OR external location
-				if( directoryExists(root & "/" & foundPaths & thisFolder)
+				if( !find( ':', foundPaths) and !find( ':', thisFolder ) and directoryExists(root & "/" & foundPaths & thisFolder)
 					OR
 				    ( len(extRoot) AND directoryExists(extRoot & "/" & foundPaths & thisFolder) )
 				){


bdw429s:
@lmajano Looks like your cachebox script changes have broken the interfaces in the build.

jclausen:
I'm curious on the use case with this one, since Windows won't allow a file to be saved with a colon in it, in the first place.  It seems like any error thrown from this method is a case of bad input and, in that case, the error should be thrown.  

If we're going to perform a check for reserved characters for that OS, shouldn't we check for them all, then? Since Unix allows these characters, technically, wouldn't we need an OS check?

bdw429s:
@jclausen If linux allows a colon then than would be an issue for sure,.  A more specific check would be to check and see if the incoming back was already absolute.  Which is why I had also suggested just reversing the checks so the absolute file check would happen first.

bdw429s:
@angel-chrystian based on Jon's feedback, would you like to try the other idea of reversing the order of the checks?

angel-chrystian:
@bdw429s I tested in the sandbox with the code like this
```
		//Check 1: Absolute Path
		if( fileExists( arguments.pathToCheck ) ){
			foundPath = arguments.pathToCheck;
		}
		//Check 2: Inside of App Root
		else if ( fileExists( variables.appRootPath & arguments.pathToCheck) ){
			foundPath = variables.appRootPath & arguments.pathToCheck;
		}
		//Check 3: Expand the Path
		else if( fileExists( expandPath( arguments.pathToCheck ) ) ){
			foundPath = expandPath( arguments.pathToCheck );
		}
```
But I get this error, on java.io.FilePermission
> Message: Event handler exception. Detail: An exception occurred while invoking an event handler method from Application.cfc. The method name is: onApplicationStart. access denied ("java.io.FilePermission" "\includes\i18n\main_es_MX.properties" "read") Raw Trace: at cfController2ecfc850527843$funcLOCATEFILEPATH.runFunction(D:\home\verdekan.com\wwwroot\coldbox\system\web\Controller.cfc:798) Tag Context: D:\home\verdekan.com\wwwroot\coldbox\system\web\Controller.cfc Raw Trace: at cfResourceService2ecfc1163844644$funcGETRESOURCEBUNDLE.runFunction(D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\models\ResourceService.cfc:205) Tag Context: D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\models\ResourceService.cfc Raw Trace: at cfResourceService2ecfc1163844644$funcLOADBUNDLE.runFunction(D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\models\ResourceService.cfc:100) Tag Context: D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\models\ResourceService.cfc Raw Trace: at cfi18n2ecfc937522341$funcCONFIGURE.runFunction(D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\models\i18n.cfc:49) Tag Context: D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\models\i18n.cfc Raw Trace: at cfModuleConfig2ecfc297281408$funcAFTERCONFIGURATIONLOAD.runFunction(D:\home\verdekan.com\wwwroot\modules\cb-bundle\cborm\modules\cbvalidation\modules\cbi18n\ModuleConfig.cfc:124)

lmajano:
Will start revising

lmajano:
Ok, so unix based OSs do allow for `:` to exist in the filenames.  So this really won't do anything:

```
luis:majano.txt
```

lmajano:
I am at a loss on how to address this folks.  Thoughts? @bdw429s @angel-chrystian

bdw429s:
@lmajano Honestly, I don't know if there is a good solution to this.  The short answer is ColdBox can't do "guessing" of where a file may be located if Sandbox security is going to step in and blow up in those cases.  It would have been best if ColdFusion simply returned false for fileexists checks in those instances.  Unless there's a way for ColdBox to know what parts of the hard drive are "off-limits" then we're sort of screwed and unable to look in our different convention locations.  

The only thing I can think of are to enable some sort of "strict" mode in ColdBox where ALL paths are forced to be absolute and don't do any guessing.

lmajano:
I will close this then, until we can think of better and non-verbose ways to tackle it.  Thanks for everything @angel-chrystian . I apologize for not being able to merge this.

PR #404, opened 2019-04-08
Load interceptors from config `onConfigurationLoad`

elpete:
Try loading interceptors onConfigurationLoad also to catch interceptors that relied on custom interception points created by modules

### DIFF
--- a/system/web/services/InterceptorService.cfc
+++ b/system/web/services/InterceptorService.cfc
@@ -70,6 +70,12 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
 		registerInterceptor(
 			interceptorObject 	= controller.getSetting( 'coldboxConfig' ),
 			interceptorName 	= "coldboxConfig"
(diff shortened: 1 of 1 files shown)


lmajano:
@elpete thanks for this pull request.  I have been thinking long and hard on how to achieve this since basically the chicken and the egg issue is difficult.

However, I am a litte hesitant on the approach because basically, we re-configure and re-create ALL interceptors again, basically twice.  Seems like a performance hit.  Thoughts on alternatives?

elpete:
You could cache an array of function names not matching interception points during the initial registration and then compare those names to the list `afterConfigurationLoad`.  If any match we could re-configure and re-register then.

lmajano:
Thanks @elpete I manually merged it as I updated some snapshot code.

https://github.com/ColdBox/coldbox-platform/commit/8b92daf50abf9afc0d097efabcb774a15b49a1ef

PR #407, opened 2019-04-18
Make default baseURL domain and protocol agnostic

jclausen:
This default causes more problems than it solves, IMHO, and complicates the process of using framework routes for Docker healthchecks.  

I'm opening this PR because I would like to see the hostname and protocol used only if a `pathInfoProvider` explicitly dictates to use it.

### DIFF
--- a/system/web/routing/Router.cfc
+++ b/system/web/routing/Router.cfc
@@ -144,9 +144,9 @@ component accessors="true" extends="coldbox.system.FrameworkSupertype" threadsaf
 		variables.validExtensions = variables.VALID_EXTENSIONS;
 		// Base Routing URL, defaults to the domain and app mapping defined by the routing services
 		if( len( controller.getSetting( "RoutingAppMapping" ) ) lte 1 ){
-			variables.baseURL = "http://#cgi.HTTP_HOST#";
+			variables.baseURL = "";
 		} else {
-			variables.baseURL = "http://#cgi.HTTP_HOST##controller.getSetting( "RoutingAppMapping" )#";
+			variables.baseURL = "#controller.getSetting( "RoutingAppMapping" )#";
 		}
 		// Are full rewrites enabled
 		variables.fullRewrites = false;


lmajano:
@jclausen what would be the side effects of this change, willit break compatiblity?  @bdw429s thoughts?

bdw429s:
I would expect this to affect anyone using `event.buldlink()` to generate an external link to their site.  For example, if you were templating an E-mail and needed the full link.  

@jclausen Can you elaborate on your use case.  Perhaps a new flag such as 
```js
event.buildLink( event='health.check', relative=true );
```

jclausen:
@lmajano :  The major side effect on this one is going to be within emails, if they use `event.buildLink`.  There may be some other areas where the expectation would be for a qualified URL and now it will be a  path.

jclausen:
Jinx with Brad! :) 

@bdw429s :  A big one is template caching. There's no where in your HTML that you should ever need an external link to have the full domain and protocol.  This also applies to static assets.

If you have layout caching turned on, for example, and you have a link you've built with `event.buildLink`, currently you're going to have links with whatever protocol and domain the page was accessed when the cache was created. Which could be a different link when the cache gets re-built, etc.

Healthchecks are also an issue, because the framework only sets that BaseURL once, by default, which means that if your healthcheck hits on 127.0.0.1, then that becomes the BaseURL domain for all `event.buildLink` calls until the framework reinitializes.

jclausen:
I'm open to the `relative` argument to `buildLink`, but I think it should default to `true`.

bdw429s:
>  framework only sets that BaseURL once, by default, which means that if your healthcheck hits on 127.0.0.1, then that becomes the BaseURL domain for all event.buildLink calls until the framework reinitializes. 

I actually thought @lmajano fixed this in ColdBox 5 due to all the troubles it caused by caching the base URL.  @lmajano Can you respond to that?

jclausen:
My understanding was that `SESBaseURL` argument to the RequestContext was provided by the Router ( https://github.com/ColdBox/coldbox-platform/blob/development/system/web/context/RequestContext.cfc#L236 ).  If it's not, any longer, then it almost makes no sense for the Router to even deal with that, at all, which actually makes the case for this PR - a singleton should never lock in the domain/port/protocol

lmajano:
THoughts on this aga

> > framework only sets that BaseURL once, by default, which means that if your healthcheck hits on 127.0.0.1, then that becomes the BaseURL domain for all event.buildLink calls until the framework reinitializes.
> 
> I actually thought @lmajano fixed this in ColdBox 5 due to all the troubles it caused by caching the base URL. @lmajano Can you respond to that?

That is correct, the first one wins.

bdw429s:
@jclausen It seems @lmajano has forgotten about this ticket which explicitly reversed the behavior of previous ColdBox versions:

https://ortussolutions.atlassian.net/browse/COLDBOX-753

Unless you turn it off, ColdBox will reset the base URL in the request context on every request here

https://github.com/ColdBox/coldbox-platform/blob/development/system/web/services/RoutingService.cfc#L178-L186

Feature defaults to "on" here:

https://github.com/ColdBox/coldbox-platform/blob/development/system/web/routing/Router.cfc#L153

So basically, ColdBox now behaves as though you always have this module installed, just without the added feature of being able to force SSL all the time.

https://www.forgebox.io/view/ses-on-request

jclausen:
I think we can close this one out then, given Coldbox has gone in a different direction in handling this.

PR #408, opened 2019-04-24
Update RequestContext.cfc getFullUrl() to include port number

homestar9:
Currently `getHTMLBaseURL()` returns the port number if it is displayed in the address bar. It seems like `getFullURL()` should also include the port number if displayed as well.  Having the port displayed is especially important if you have a `<base>` tag in your layout and use anchor links (e.g. `<a href="#event.getFullURL()###anchor"...`.  

current:
`event.getHTMLBaseURL()` = `http://127.0.0.1:52606/`
`event.getFullURL()` = `http://127.0.0.1/handler/action/`

proposed:
`event.getHTMLBaseURL()` = `http://127.0.0.1:52606/`
`event.getFullURL()` = `http://127.0.0.1:52606/handler/action/`

### DIFF
--- a/system/web/context/RequestContext.cfc
+++ b/system/web/context/RequestContext.cfc
@@ -968,6 +968,7 @@ component serializable=false accessors="true"{
         return arrayToList( [
             isSSL() ? "https://" : "http://",
             CGI.SERVER_NAME,
+			listLen( variables.SESBaseURL, ':' ) == 3 ? ":" & CGI.SERVER_PORT : "",
             isSES() ? "" : "/index.cfm",
             CGI.PATH_INFO,
             CGI.QUERY_STRING != "" && CGI.PATH_INFO == "" ? "/" : "",
--- a/system/web/context/RequestContext.cfc
+++ b/system/web/context/RequestContext.cfc
@@ -968,7 +968,7 @@ component serializable=false accessors="true"{
         return arrayToList( [
             isSSL() ? "https://" : "http://",
             CGI.SERVER_NAME,
-			listLen( variables.SESBaseURL, ':' ) == 3 ? ":" & CGI.SERVER_PORT : "",
+			listFind( "80,443", cgi.server_port ) ? "" : cgi.server_port,
             isSES() ? "" : "/index.cfm",
             CGI.PATH_INFO,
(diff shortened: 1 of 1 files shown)


bdw429s:
Will this omit the port when it's `80` or `443`?

homestar9:
It will only show the port if it is present in `variables.SESBaseURL`.  If the port is omitted there, it should be omitted in the result of `getFullURL()`

system/web/context/RequestContext.cfc
```diff
@@ -968,6 +968,7 @@ component serializable=false accessors="true"{
         return arrayToList( [
             isSSL() ? "https://" : "http://",
             CGI.SERVER_NAME,
+			listLen( variables.SESBaseURL, ':' ) == 3 ? ":" & CGI.SERVER_PORT : "",
```
lmajano: Why 3?  Shouldn't it be if it finds two `:` in the URL?

lmajano: Also, modules could be leverages that use the `:` notation. SO it could basically give you a false positive.

I think this should determine the port qualifier to be in certain numbers just to be safe.

```
listFind( "80,443", cgi.server_port ) ? "" : cgi.server_port
```

homestar9: Good point about modules using the : notation.

The solution of looking for ports 80 or 443 could work as well since those are typically omitted.  Another idea would be to use a regex to check for the existence of a port number in the url: `(https?://.*):(\d*)\/?(.*)` and if true, show the port number.
https://www.regextester.com/99331

homestar9: Darn, that regex will still false-positive on the following url: http://www.domain.com/test/:module/.  I'll see if I can find one that will match just the port if present.

lmajano: I think it would be best to jsut see if the server port is within the list

homestar9: Using a predefined list sounds good to me.  While technically http://127.0.0.1:80/ and https://127.0.0.1:443/ are perfectly valid, there's no real reason to include those ports in the URL.

This fix should address the issues impacting people using non-standard ports (like during development)  http://127.0.0.1:52606/

PR #409, opened 2019-05-13
Initial pass at merging HTTP verbs for the same route

elpete:
Currently, the following doesn't work as expected:

```
router.post( "photos/", "photos.create" );
router.get( "photos/", "photos.index" );
```

You would expect then you could either `GET` or `POST` to `photos/`.  This is not the case.  Instead this creates two routes.  If you tried to `GET /photos` you would match the route that only has a `POST` verb and it would give you an error of an invalid HTTP verb.

This PR attempts to merge routes when a route with a matching pattern already exists in the routing table.

### DIFF
--- a/system/web/routing/Router.cfc
+++ b/system/web/routing/Router.cfc
@@ -631,7 +631,7 @@ component accessors="true" extends="coldbox.system.FrameworkSupertype" threadsaf
 
 	        // update, delete and show routes
 	        actionSet = filterRouteActions(
-	        	{ PUT = "update", PATCH = "update", POST = "create", DELETE = "delete", GET = "show" },
+	        	{ PUT = "update", PATCH = "update", DELETE = "delete", GET = "show" },
 	        	arguments.only,
 	        	arguments.except
 	        );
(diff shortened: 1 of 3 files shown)


system/web/routing/Router.cfc
```diff
@@ -631,7 +631,7 @@ component accessors="true" extends="coldbox.system.FrameworkSupertype" threadsaf
 
 	        // update, delete and show routes
 	        actionSet = filterRouteActions(
-	        	{ PUT = "update", PATCH = "update", POST = "create", DELETE = "delete", GET = "show" },
```
lmajano: Why did you remove `create`, this will break all resources that use a `POST` for creation though?

elpete: You don’t `POST` to the `/:id` route. You post to the base route.

elpete: That `POST` was a duplicate.

lmajano: Really, where is it defined, can't see it?

elpete: Further down: https://github.com/ColdBox/coldbox-platform/blob/f3c873dec5169238b3653a1dcd4bb9d4298588c0/system/web/routing/Router.cfc#L649

elpete:
Can you check this again, @lmajano?  I responded days ago.

lmajano:
Revising this again. It looks fine but I am concerned we are only focused on one use case
`verb( pattern, event)`.  What if we have concatenation to handler/action, where `event` does not exist in the route, but it is split?

elpete:
Give me a list of all of the possibilities, we’ll add tests, and make it work.

lmajano:
Yes sir!

bdw429s:
@lmajano Waiting on you for this one.

PR #410, opened 2019-06-03
Fix getSetting typo

elpete:


lmajano:
COLDBOX-786 #resolve

HTMLHelper typo on getSetting call via elixir path

PR #411, opened 2019-06-21
Fix default bug report to show SQL error detail for adobe SQL exceptions

ryanalbrecht:
Adobe coldfusion returns false when executing isStruct() against an exception struct. This caused the default coldbox error template to not show the SQL exception details.

Added in another statement to check the underlying class name of the exception struct to determine if it is a database exception.


lmajano:
Thanks for this!  We really appreciate it!

PR #414, opened 2019-06-25
COLDBOX-799 handler bean caching

DominicWatson:
This change makes dramatic improvements to our applications. Checking the logic, it appears to me that we do not need to create a new instance of a handler bean every time we run a handler action, rather, we should create a single, persisted instance for each handler action and only read the metadata of a handler CFC/method once.

I could be wrong, of course! So would be good to get a thorough check on this. I have tested with concurrency in Preside applications with no side effects.


system/web/context/EventHandlerBean.cfc
```diff
@@ -61,57 +66,62 @@ component accessors="true"{
...
-		return this; 
+		return this;
 	}
-	function setHandlerMetadata( required handlerMetadata ){ 
+	function setHandlerMetadata( required handlerMetadata ){
 		variables.handlerMetadata = arguments.handlerMetadata;
-		return this; 
+		return this;
+	}
+	function setMetadataLoaded( required metadataLoaded ){
```
lmajano: why add this setter if accessors is already on?

system/web/services/HandlerService.cfc
```diff
@@ -224,10 +234,20 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
...
+		}
+
 		var handlersList 			= variables.registeredHandlers;
 		var handlersExternalList 	= variables.registeredExternalHandlers;
-		var oHandlerBean 			= new coldbox.system.web.context.EventHandlerBean( variables.handlersInvocationPath );
 		var moduleSettings 			= variables.modules;
+		var oHandlerBean 			= new coldbox.system.web.context.EventHandlerBean( variables.handlersInvocationPath );
+
+		// put bean in cache if enabled
+		if ( variables.handlerCaching ) {
```
lmajano: I will modify this slightly, because we do not want to cache ALL event handler beans because the actual event could be invalid.  This could case us to create and cache tons of unecessary beans that will never be used.

system/web/context/EventHandlerBean.cfc
DominicWatson: Just in line with all the others you have is all. to make chainable. Not necessary.

system/web/services/HandlerService.cfc
DominicWatson: Good point, I think there is also a thread safety issue there. Let me refactor that now.

lmajano:
Dom I have merged all your changes, but for some reason I had to do it manually.  Anyways, the changes are incorporated and I took advantage in creating all bdd tests for the different situations.  Thanks for the amazing work you do in helping us out!  We really appreciate you!

lmajano:
Merged Manually: https://github.com/ColdBox/coldbox-platform/commit/e9992a30e4b5125ba364aeb6fe13361d958441ce

DominicWatson:
Nice one :) and likewise - appreciate the tireless effort you guys put in!

PR #417, opened 2019-07-31
COLDBOX-807 local var scoping when using IsNull()

DominicWatson:
Checked whole codebase for IsNull( bareVar ) and found a few instances. Having been using `IsNull( local.someVar )` in Preside for some time now due to the strange bugs this causes. 

You can test for the problem this fixes with something like:

```cfc
// handlers/test.cfc
component {
    property name="cache" inject="cachebox:default";

    function index() {
        var fromCache = cache.getOrSet( objectKey=CreateUUId(), producer=function(){
            return "From Producer";
        } );
       event.renderData( data={ fromCache=fromCache }, type="json" );
    }
}
```

Hit that in a browser and see 'From Producer' in the result. Then add `?target=test` and see the result change.

**Whitespace notes**: Looks like `/system/EventHandler.cfc` has some strange line br chars (in the core repo). Hence the diff being all on one line :o. I also have trim whitespace on save set in my editor, so apologies for extra whitespace changes.


bdw429s:
@DominicWatson This all looks really good to me, but can you tell me what's going on with the `system/EventHandler.cfc` file?

bdw429s:
@DominicWatson Did you see my comment above about the weirdness in the diff?

DominicWatson:
Yeah, check out the file in GitHub browser in the coldbox repo. I commented
on that in the PR.

bdw429s:
@DominicWatson Sorry, I didn't see that original note.  I just opened the file up in the online editor and re-saved and it seems to have straightened out the line breaks which I think was the original issue.

DominicWatson:
Nice one :)

PR #420, opened 2019-08-19
Feature/cachebox 57 lucee provider config fix

sjdaniels:


system/cache/providers/CFProvider.cfc
```diff
@@ -212,10 +212,13 @@ component
      * Get a structure of all the keys in the cache with their appropriate metadata structures. This is used to build the reporting.[keyX->[metadataStructure]]
      */
     struct function getStoreMetadataReport(){
-		return getKeys()
-			.map( function( item ){
-				return getCachedObjectMetadata( item );
+		var result = {};
+		getKeys()
```
lmajano: hmm, what is the difference of using map() vs this approach? Map does exactly the same

system/cache/providers/CFProvider.cfc
```diff
@@ -538,14 +541,7 @@ component
...
 		 // Add in settings not discovered
-		structAppend( variables.configuration, variables.DEFAULTS );
-		// Validate configuration values, if they don't exist, then default them to DEFAULTS
-		for( var key in variables.DEFAULTS ){
-			if( NOT len( variables.configuration[ key ] ) ){
-				variables.configuration[ key ] = variables.DEFAULTS[ key ];
-			}
-		}
-
+		structAppend( variables.configuration, variables.DEFAULTS, false );
```
lmajano: Good point, right now it DOES override it.

system/cache/providers/lucee-lib/LuceeStats.cfc
```diff
@@ -71,7 +71,7 @@ component implements="coldbox.system.cache.util.IStats" accessors="true"{
 	/**
 	 * Get the total cache's hits
 	 */
-	numeric function getHits();
+	numeric function getHits() {
```
lmajano: Good catch

system/cache/providers/LuceeProvider.cfc
```diff
@@ -111,10 +111,13 @@ component
      * Get a structure of all the keys in the cache with their appropriate metadata structures. This is used to build the reporting.[keyX->[metadataStructure]]
      */
     struct function getStoreMetadataReport(){
-		return getKeys()
-			.map( function( item ){
-				return getCachedObjectMetadata( item );
+		var result = {};
```
lmajano: Same as above, the `map()` function does the same.

system/cache/providers/CFProvider.cfc
sjdaniels: array().map() returns an array. This function is supposed to return a struct.

system/cache/providers/LuceeProvider.cfc
sjdaniels: See above.

system/cache/providers/CFProvider.cfc
lmajano: Ohhhhhhhhh, ok, then let’s switch it up to a `reduce()` function then, so we can eat our cake 🍰

sjdaniels: OK. All set.

PR #422, opened 2019-08-21
Fix routing bug from stacking multiple verb methods

elpete:


elpete:
Resolves https://ortussolutions.atlassian.net/browse/COLDBOX-812

system/web/services/RoutingService.cfc
```diff
@@ -315,7 +315,11 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
 				if( log.canDebug() ){
 					log.debug( "Invalid HTTP Method detected: #httpMethod#", routeResults.route );
 				}
-			}
+            }
+
+            if ( isStruct( routeResults.route.action ) && structIsEmpty( routeResults.route.action ) ) {
```
lmajano: What is the purpose of this code?

elpete: The code below only knows how to deal with empty string actions, not empty structs.  The changes to merging routes can introduce an empty struct.  This code resets an empty struct to an empty string so it gets handled by the wildcard or not found routing below.

lmajano: Wouldn't it be best if on line 325 we just add the second check instead of a reset.  I think we should just check if the action is a struct and if it is empty.

lmajano: @elpete

lmajano:
@elpete all tests are failing.

lmajano:
I wanted to merge, but all tests fail

elpete:
I'm happy to fix once you get `development` working again.

elpete:
@lmajano It passes.  Merge it. ;-)

PR #426, opened 2019-09-10
Remote POST from resources hint

GaryStanton:


GaryStanton:
^ By which I mean _remove_, obviously...

PR #428, opened 2019-09-15
LOGBOX-35 always allow spawning of log processor thread when engine supports it

DominicWatson:
This prevents issues where the log processing thread never spawns, or does not spawn in a timely manner, when logging in background threads. This problem can lead to memory usage issues because the log queue is held in heap and cannot be collected.

### DIFF
--- a/system/core/util/Util.cfc
+++ b/system/core/util/Util.cfc
@@ -10,6 +10,13 @@ Description :
 ----------------------------------------------------------------------->
 <cfcomponent output="false" hint="The main ColdBox utility library filled with lots of nice goodies.">
 
+	<!--- constructor --->
+	<cffunction name="init" access="public" returntype="any" output="false">
+		<cfscript>
+			checkThreadWithinThreadSupport();
+		</cfscript>
+	</cffunction>
+
 	<!--- getMixerUtil --->
     <cffunction name="getMixerUtil" output="false" access="public" returntype="any" hint="Get the mixer utility" doc_generic="coldbox.system.core.dynamic.MixerUtil">
     	<cfscript>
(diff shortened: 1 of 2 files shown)


system/logging/appenders/FileAppender.cfc
```diff
@@ -306,8 +306,10 @@ component accessors="true" extends="coldbox.system.logging.AbstractAppender"{
 	 * @message The target message
 	 */
 	private FileAppender function append( required message ){
-		// If we are not in a thread, then start the log listener, else queue it
-		if( !getUtil().inThread() ){
+		var util = getUtil();
+
+		// if thread within thread not supported, check we're not already in child thread
+		if( util.threadWithinThreadSupported() || !util.inThread() ){
```
lmajano: Actually, out of curiosity, what would happen if we just don't do the check? Will the CFML engines create the `cfthread` or just thrown an exception?

system/core/util/Util.cfc
```diff
@@ -94,6 +101,29 @@ Description :
 		</cfscript>
 	</cffunction>
 
+	<!--- threadWithinThreadSupported --->
+	<cffunction name="threadWithinThreadSupported" access="public" returntype="boolean" hint="I return whether or not the current engine allows spawning threads from child threads" output="false">
+		<cfreturn variables.threadWithinThreadSupported />
+	</cffunction>
+
+	<!--- checkThreadWithinThreadSupport --->
+	<cffunction name="checkThreadWithinThreadSupport" access="private" returntype="any" output="false" hint="A little test logic to check for thread within thread support">
```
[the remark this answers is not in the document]
DominicWatson: > Couldn't we just do this by engine detection, or is it a specific version of Lucee that introduced this?

My thinking was that, should ACF not do this now, but change it in the future, it would just work. It is a specific version of Lucee that fixed it.

system/logging/appenders/FileAppender.cfc
DominicWatson: Lucee, prior to the fix, would throw an exception before creating thread (no thread created). 
Open BlueDragon would always allow it
ACF - ? I dunno. 

But yes, the thread tag could just be wrapped in a try catch. But that is perhaps less efficient (would need to perf test that for memory usage, etc)?

lmajano:
@DominicWatson I have attacked this in a different approach by leveraging a native Java Runnable.  So far all my tests pass on adobe and lucee on this. Check this out: https://github.com/ColdBox/coldbox-platform/commit/643e5814b65ed7853823edfec617bcf118bd56c0

I will also try to use this approach on other useful utilties within ColdBox and maybe open it via some helper methods

lmajano:
Thanks for this @DominicWatson It was inspiring to add runnable core support into ColdBox.  So far the BE is now using this new approach.

DominicWatson:
Looks awesome!

PR #430, opened 2019-10-17
Add a whileSwapped test helper to swap out wirebox mappings inside a callback

elpete:
Used like so:

```cfc
component extends="coldbox.system.testing.BaseTestCase" {

    function run() {
        describe( "Currency API", function() {
            it( "can convert an amount from one currency to another", function() {
                var currencyApiStub = createStub()
                    .$( "getExchangeRate" )
                    .$args( "EUR", "USD" )
                    .$results( 1.11 );
                whileSwapped( { "currencyAPI": currencyApiStub }, function() {
                    var event = get( "/currency/convert", {
                        "amount": 10.00,
                        "fromCurrency": "EUR",
                        "toCurrency": "USD"
                    } );
                    expect( event.getStatusCode() ).toBe( 200 );
                    expect( event.getRenderData().data ).toBe( 11.10 );
                } );
            } );
        } );
    }

}
```

### DIFF
--- a/system/testing/BaseTestCase.cfc
+++ b/system/testing/BaseTestCase.cfc
@@ -835,6 +835,45 @@ component extends="testbox.system.compat.framework.TestCase"  accessors="true"{
         } else {
             return getPageContext().getResponse().getResponse();
         }
-	}
+    }
+
+    /**
+     * Swaps out WireBox mappings for corresponding mocks during a callback.
+     *
+     * @mappings                A struct of mappings to swap the given WireBox mapping
+     *                          for the given value, i.e. { "apiClient" = mockApiClient }
+     * @callback                The callback to execute with the swapped mappings.
+     * @verifyMappingsExists    Flag to first verify that the mapping exists before swapping it.
+     *                          Useful to ensure you have set the mapping for your application
+     *                          and not just your tests.
+     */
+    public void function whileSwapped(
+        struct mappings = {},
+        any callback,
+        boolean verifyMappingExists = true
+    ) {
+        var binder = getWireBox().getBinder();
+        var originalMappings = {};
+        mappings.each( function( mapping, component ) {
+            var exists = binder.mappingExists( mapping );
+            if ( verifyMappingExists && ! exists ) {
+                expect( exists ).toBeTrue( "No mapping [#mapping#] already configured in WireBox." );
+            }
+            if ( exists ) {
+                originalMappings[ mapping ] = binder.getMapping( mapping );
+            }
+            binder.map( alias = mapping, force = true ).toValue( component );
+        } );
+
+        try {
+            callback();
+        } catch ( any e ) {
+            rethrow;
+        } finally {
+            originalMappings.each( function( mapping, component ) {
+                binder.setMapping( mapping, component );
+            } );
+        }
+    }
 
 }
--- a/tests/specs/integration/IntegrationTestSpec.cfc
+++ b/tests/specs/integration/IntegrationTestSpec.cfc
@@ -1,0 +1,71 @@
+component extends="coldbox.system.testing.BaseTestCase" appMapping="/cbTestHarness" {
+
+    function run() {
+        describe( "coldbox integration test helpers", function() {
+            describe( "whileSwapped WireBox helper", function() {
+                it( "can swap out a WireBox mapping inside a closure", function() {
+                    var mappingName = "test-mapping";
+                    var originalInstance = new tests.resources.Class1();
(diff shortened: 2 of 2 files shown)


bdw429s:
@elpete Is this to swap out a Wirebox mapping just for the duration of a single test? I really like this and I needed it in my integration tests the other day so i could mock something without permanently changing the loaded Coldbox app.  

I would suggest a better name however as `whileSwapped` isn't even clear that it applies to wirebox, the binder, or mappings.

bdw429s:
I'll also add that this method won't prevent a copy of your stub from getting permanently injected elsewhere in the app that lives on even after the while swapped closure is done.  In fact, based on the order of tests, it's possible the mapping in question may have already been built and cached as an injection in a previous test which means the stub may not even be used.  A workaround to this would be to wrap the injection in a provider, but that may add unnecessary overhead to the app under normal conditions.

bdw429s:
>  only swaps out the binding inside the closure, not the entire test case.

Yes, I know and my comments still stand.  

> We'd need to be able to register TestBox after test hooks inside specs for that to work.

For what to work?  I think perhaps you think I was suggesting a change to what you implemented?  I think only swapping for the life of the closure is a great idea.

> I'm fine with a different name. I thought about putting this on the WireBox's Injector object so it would be used wirebox.whileSwapped but then you have to add it just for tests.

I had the same thought.  Even if the method was on the injector, I'd want it to have a better name though.  At the end of the day, you're swapping mappings in the binder config inside of the injector so I'd expect to at least see the word "mapping" in the name.

> The problem of cached components and scope widening is interesting. 

To be clear, I didn't describe any scope widening.  

> I haven't run in to it in practice yet.

It would be as simple as injecting a service into a handler and then having several actions that use the injected service.  Let's say you have a spec that tests one action in that handler, which loads up the handler instance into ColdBox and injects the service.  Then you have another spec that attempts to swap the mapping for the service during a ColdBox event execution, except the service is already injected into the cached handler, so WireBox is not asked to build it again and the temporary mapping never gets used.

> Is there a way to invalidate any WireBox components that depend on swapped WireBox mappings? 

Not that I know of.  A CFC instance with injections could be anywhere-- a cache, the session or application scope, any of the internal ColdBox services, or inside of a deep dependency hierarchy chain.  The only bulletproof way to force it would be to tear ColdBox down and rebuild it before AND after the "swapped" closure that performs the mapping swap.

elpete:
I'd be fine with resetting WireBox when using this utility.  I'm not sure the performance impact that would have.

bdw429s:
> I'd be fine with resetting WireBox when using this utility. 

Not WireBox, but all of ColdBox

> I'm not sure the performance impact that would have.

It would be the same as leaving `this.unloadColdBox=true` but just for that spec.

lmajano:
After reviewing this, there are several issues to address.

1) This only addresses mappings that have not been instantiated at the point in time where the function is called.  So it won't work unless you completely restart the framework.
2) Why not just get the object and mock what you need and then clear the singletons


```
prepareMock( getInstance( "RantService" ) )
	.$( "getAll", [] );

... DO YOUR TESTING

// Clean the mocks
getWireBox().clearSingletons();
```

I don't see the point of hijacking mappings when you can mock any object explicitly

elpete:
I'll address these in reverse order:

> 2. Why not just get the object and mock what you need and then clear the singletons

Two reasons.  1) I may be swapping out a transient like `QueryBuilder@qb`. 2) I may want to supply my own fake object instead of mocking it.

> This only addresses mappings that have not been instantiated at the point in time where the function is called. So it won't work unless you completely restart the framework.

I guess.  My only response to this is I haven't ran in to this in practice since the BaseTestCase loads and unloads ColdBox by default.  It seems that the `getWireBox().clearSingletons()` trick would work here as well.

lmajano:
> Two reasons. 1) I may be swapping out a transient like QueryBuilder@qb. 

You can still mock a transient. Or mock anything within the call. I do this all the time.

> 2) I may want to supply my own fake object instead of mocking it.

What would be the difference? You will have to mock methods anyways

>I guess. My only response to this is I haven't ran in to this in practice since the BaseTestCase loads and unloads ColdBox by default. It seems that the getWireBox().clearSingletons() trick would work here as well.

THe issue is you will have to unload the framework and clear wirebox from app scope first, then load and then clear.  I just don't see it very practical to do all this machinery, when you can get access to any object and use mocks and spies.

elpete:
I'll close this issue then.  I'll likely keep using this in my projects though since it fits my mental model.  Thanks.

PR #431, opened 2019-10-28
Option to call Function to calculate Event Cache Timeout

gpickin:
Clients have requested the ability to set a specific timeout for Event Caching... for example midnight... currently we cannot do that. 
Lucee for one, does not allow us to use a UDF or Closure or even a variable in the function meta data, only a constant value. So we have to manually guess a timeout, 120 or 500 or something, or make it 1440 and then manually clear the caches with a scheduled task or something.
 
I propose we use metadata in the function to name the DSL of the Model, and metadata to name the method to call on the Model to calculate this timeout value.

{code:java}
cacheTimeoutUdfInstance1="apiUtils@api" 
cacheTimeoutUdfMethod1="getMinutesUntilMidnight"
{code}

If both of these metadata values are present, this would call 
`wirebox.getInstance( "apiUtils@api" ).getMinutesUntilMidnight()` 
and override the metadata for the timeout.

Pull request is attached. 
I am not sold on the names, and maybe there is a more stylish way to do it ( a shorthand maybe so we can use one variable ).

https://ortussolutions.atlassian.net/browse/COLDBOX-828

### DIFF
--- a/system/web/services/HandlerService.cfc
+++ b/system/web/services/HandlerService.cfc
@@ -183,7 +183,9 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
 				// Create caching data structure according to MD, as the cache key can be dynamic by execution.
 				var eventCachingData = {};
 				structAppend( eventCachingData, eventDictionaryEntry, true );
-
+				if( len( eventCachingData.cacheTimeoutUdfInstance ) && len( eventCachingData.cacheTimeoutUdfMethod ) ){
+					eventCachingData.timeout = wirebox.getInstance( eventCachingData.cacheTimeoutUdfInstance )[ eventCachingData.cacheTimeoutUdfMethod ]();
+				}
 				// Create the Cache Key to save
 				eventCachingData.cacheKey = oEventURLFacade.buildEventKey(
 					keySuffix 		= eventDictionaryEntry.suffix,
@@ -583,6 +585,8 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
 			"cacheable" 		    = false,
 			"timeout" 		        = "",
 			"lastAccessTimeout" 	= "",
+			"cacheTimeoutUdfInstance" = "",
+			"cacheTimeoutUdfMethod" = "",
 			"cacheKey"  		    = "",
 			"suffix" 			    = "",
 			"provider"				= "template"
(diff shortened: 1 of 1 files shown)


bdw429s:
@gpickin All the Adobe engines are broken in the build.

Two questions:

1) What if I just want to use a UDF local to the handler I'm caching in?
2) Can you provide some more examples or the custom times clients have needed in addition to "midnight"?  I'm curious if a simpler solution that introduced a DSL of reserved words could be used to signal specific predetermined calculations like "midnight".

elpete:
At a previous job we had multiple database jobs that ran at different times during the night: 12 AM, 1 AM, 3 AM, etc.  I think the right approach is to allow for a UDF, both local and from another CFC, so we can be as flexible as possible.

gpickin:
I am leaning toward a simple attributes for a calculated timeout, based on a time... and a udf with optional wirebox dsl for a udf in a different file.

gpickin:
I wrote an interceptor to do this, so we dont have to mess with the core. Although I'd want someone to tell me how to improve the process of getting the metadata of the event action.
Right now I'm having to do this, to get a event handler bean with the meta data in there so I can get `cacheOverrideTimeout` meta value, so see if we need to override the timeout.

```
function preRender( event, interceptData, buffer, rc, prc ) {
    var cacheMeta = prc.cbox_eventCacheableEntry;
    if( structKeyExists( cacheMeta, "cacheable" ) && cacheMeta.cacheable ){
    	    var cacheOverrideTimeout = getCacheOverrideTimeout( arguments.event );
    	    if( len( cacheOverrideTimeout ) and cacheOverrideTimeout == 'midnight' ){
    		cacheMeta[ "timeout" ] = getMinutesUntilMidnight();
	    }
    }
}
private function getCacheOverrideTimeout( event ) {
    var ehbean = controller.getHandlerService().getHandlerBean( event.getCurrentEvent() );
    var oEventHandler = controller.getHandlerService().newHandler( ehBean.getRunnable() );
    ehBean.setActionMetadata( oEventHandler._actionMetadata( ehBean.getMethod() ) );
    ehBean.getActionMetadata( "cacheOverrideTimeout", "" );
}
```

lmajano:
I am divided on this, in terms of approach. As it stands it will not work on either approaches you have presented.

As it stands, timeouts are **fixed** as they are compiled into the metadata, and we know that metadata can't be dynamic on any engine.  So to provide a mechanism where the framework can call something to get a calculated/dynamic timeout it has to be done by dsl or convention.  Understand that! ✔️ 

I would prefer to introduce a convention to call a UDF in the same handler to calculate the timeout instead of two more annotations for model and method.  from that function you can call whatever you like instead of adding complexity to the annotation.  

Another problem with this pull request is also that the metadata **inspection** is different than metadata **usage**.  Meaning that the annotations are read in and executed at the time of metadata cache creation is NOT the same as when that metadata is used to register the cacheable entry.

Meaning that ONLY on first event executed will the timeout be calculated not on EVERY permutation of the cache entries.  This will create all entries expiring incorrectly or expiring only on the FIRST calculated timeout.

So this is not as easy as it sounds because not only do you have to store the metadata, you have to execute the metadata at the time the cacheable entry gets generated (line 185 on Handler Service)

```java
// Get event action caching metadata: HERE IS WHERE THE METADATA IS READ AND CACHED
var eventDictionaryEntry = getEventCachingMetadata( arguments.ehBean, oEventHandler );

// Do we need to cache this event's output after it executes??
if ( eventDictionaryEntry.cacheable ){
	// Create caching data structure according to MD, as the cache key can be dynamic by execution.
	var eventCachingData = {};
	structAppend( eventCachingData, eventDictionaryEntry, true );

// HERE IS WHERE WE WOUDL NEED TO CALCULATE THE TIMEOUT

	// Create the Cache Key to save
	eventCachingData.cacheKey = oEventURLFacade.buildEventKey(
		keySuffix 		= eventDictionaryEntry.suffix,
		targetEvent 	= arguments.ehBean.getFullEvent(),
		targetContext 	= oRequestContext
	);

	// Event is cacheable and we need to flag it so the Renderer caches it
	oRequestContext.setEventCacheableEntry( eventCachingData );

} //end if md says that this event is cacheable
```

lmajano:
@gpickin any updates?

lmajano:
I never got any updates on this, so I need updates, Or I will close as stale

PR #432, opened 2019-11-05
Implicit variable access cleanup.

Daemach:
I scoped a number of variables that were being accessed implicitly.  This led to a 25% performance improvement as some of these variables were being accessed over 1000 times in a single page load.  There is more to do, but I took care a lot of the low hanging fruit.

### DIFF
--- a/system/FrameworkSupertype.cfc
+++ b/system/FrameworkSupertype.cfc
@@ -503,7 +503,7 @@ component serializable="false" accessors="true"{
 	* @return FrameworkSuperType
 	*/
 	any function includeUDF( required udflibrary ){
-		var appMapping		= controller.getSetting( "AppMapping" );
+		var appMapping		= variables.controller.getSetting( "AppMapping" );
 		var UDFFullPath 	= expandPath( arguments.udflibrary );
 		var UDFRelativePath = expandPath( "/" & appMapping & "/" & arguments.udflibrary );
 
--- a/system/core/util/Util.cfc
+++ b/system/core/util/Util.cfc
@@ -283,7 +283,7 @@ Description :
 		<cfset var loc = {}>
 
 		<!--- First time through, get metaData of component.  --->
-		<cfif structIsEmpty( md )>
+		<cfif structIsEmpty( arguments.md )>
 			<cfif isObject( component )>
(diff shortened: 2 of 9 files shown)


lmajano:
Hmm, why did you close it?

Daemach:
I closed it because it broke ACF.  I tried for about 4 hours to get an
environment set up in which I could try to figure out what was wrong, but
kept running into problems.  I need to get back to work, unfortunately, but
IMHO, scoping variables is something you should look into.  In lucee my
page load times went from 460ms to 311ms which is a ~30% performance
improvement.  30% is not insignificant, especially given the relatively
small time investment.  To do this quickly I would recommend the following
steps:

   1. Turn on debugging info including implicit variable access:
   https://www.screencast.com/t/u6JXCMTJ2bJ
   2. Load a page with a lot of processing, scroll to the implicit variable
   access section, copy the entire table then paste into a blank spreadsheet.
   Sort by the count column desc: https://www.screencast.com/t/MLtN4DPj6aa
   The Scope column tells you where the engine finally found the variable, so
   that's what the scope should be.  The count column tells you how many times
   the engine had to search ALL of the scopes to find it - reducing higher
   counts are the low hanging fruit and largest performance gains.  File and
   line numbers make it easy to fix.  On the page I was testing, there were 2
   variables that were accessed over 1000 times in a single page load.
   Scoping just those 2 variables made for a huge improvement.
   3. Once you're done, try a couple fwreinits too.

Here's a zip of the bleeding edge build ( with Brad's fixes to those
sessioncluster settings )
https://www.dropbox.com/s/w9uyaosd2coo26p/coldbox.zip?dl=0 with my fixes.
This passes the lucee builds but not ACF.  I'm sure by using a merge tool
and your experience you can figure out why much more quickly than I can.
Hopefully it saves you some time.

lmajano:
Wow, thanks so much @Daemach . We really appreciate all the help and hard work.

Daemach:
Hi Luis -

Were you able to get these changes worked in?  The difference between
running the base code and the code with variables scoped is noticeable and
I have a scheduled task that could really use the extra speed.

lmajano:
I am on it sir!!!

PR #433, opened 2019-12-11
Singleton renderer

bdw429s:
This allows the renderer to be a singleton with NO thread safety issues.  CFModule is used to render views which fully encapsulates their variables scope inside the module call.  Request-specific instance data has been moved from the renderer.  In an example page that calls renderView() 3,000 times, the overall page time dropped from 9 seconds to 700 ms with my changes.   The full test suite is passing on Lucee and Adobe.  This pull also contains fixes necessary for the test suite to even work on Adobe without crashing, including changes to cfthread to avoid messy and costly duplication of CFC instances.

### DIFF
--- a/.cfconfig.json
+++ b/.cfconfig.json
@@ -26,5 +26,6 @@
     },
     "debuggingEnabled": false,
     "ajaxDebugWindowEnabled": false,
-    "debuggingReportExecutionTimes": false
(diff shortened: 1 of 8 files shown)


bdw429s:
Here is the related ticket
https://ortussolutions.atlassian.net/browse/COLDBOX-749

PR #434, opened 2019-12-12
Fix for  CACHEBOX-59 

eppinger:


lmajano:
Thanks @eppinger and a merry christmas to you!

PR #436, opened 2020-01-22
Use relocate and setNextEvent status codes in getStatusCode

elpete:


lmajano:
Can you expand on why this pull request?

elpete:
I needed to verify the status code for a relocate call and `getStatusCode` doesn’t ever check those values.

PR #437, opened 2020-01-31
Make withExceptionHandling useful by returning the result

elpete:


jclausen:
@elpete On adobe engines, you're going to need to use `cfconfig` to set the `errorStatusCode` to `true` in order for your tests to pass.  Out of the box, ACF returns a 200 on errors - don't ask me why.

bdw429s:
> Out of the box, ACF returns a 200 on errors - don't ask me why.

it's a security "feature" meant to prevent hackers from finding out via a script if they've caused an error on your site.  That said, I hate it because it's dumb and HTTP status codes should always represent what happened in the request.  I have seen security scanners give "false positives" just because they were able to get a 500 status code back.

system/testing/BaseTestCase.cfc
```diff
@@ -835,6 +843,32 @@ component extends="testbox.system.compat.framework.TestCase"  accessors="true"{
         } else {
             return getPageContext().getResponse().getResponse();
         }
+    }
+
+    /**
```
lmajano: @elpete can you remove this, since we talked about leveraging the exception bean approach.

lmajano: @elpete I want to merge this, but can you provide some cleanup please.

lmajano (changes requested):

elpete:
I don't remember what I was trying to do here.  We're probably better starting from scratch. :sweat_smile:

PR #438, opened 2020-02-19
Work around for adobe bug CF-4204874

ryanalbrecht:
Removed anonymous function and replaced with a UDF as a work around to adobe bug CF-4204874. This is causing a memory leak which ultimately causes my application to crash
See detail bug information here https://tracker.adobe.com/#/view/CF-4204874

### DIFF
--- a/system/logging/appenders/ConsoleAppender.cfc
+++ b/system/logging/appenders/ConsoleAppender.cfc
@@ -210,21 +210,4 @@ component accessors="true" extends="coldbox.system.logging.AbstractAppender"{
 		return this;
 	}
 
-
-	/**
-	* Wraps the provided function with a locking tag.
-	*
-	* @type The lock type. Exclusive or readonly. Defaults to exclusive
-	* @body The function to wrap
-	*/
-	private function lock( type="exclusive", body){
-		lock 	name="#getHash() & getName()#-logListener"
-				type=arguments.type
-				timeout="#variables.lockTimeout#"
-				throwOnTimeout=true	{
-			return arguments.body();
-		}
-	}
-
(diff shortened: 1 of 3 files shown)


jclausen:
@lmajano : I would recommend changing the branch target on this one to `master` and merging it in as a version patch, since this lock is used every time the log listener starts and stops.

lmajano:
Thanks guys! I will have to find a way to get a patch for 5.x since we are just working on 6 now

lmajano:
@ryanalbrecht can you send this to `master` as well?

system/logging/appenders/ConsoleAppender.cfc
```diff
@@ -221,4 +210,21 @@ component accessors="true" extends="coldbox.system.logging.AbstractAppender"{
 		return this;
 	}
 
+
```
lmajano: Since we have moved this to a UDF and most likely tons of appenders can leverage it, I would suggest you send this to the `AbstractAppender` instead so any appender can leverage the locking utility.

ryanalbrecht: Will do!

ryanalbrecht:
@lmajano 

Not 100% sure if I have done this the correct way. 

-I checked out the master branch,
-Then I used the cherry-pick command to commit the change from the development branch into master branch
-Then created a new pull request.

Is this the way you wanted it done?

lmajano:
Thanks @ryanalbrecht great job sir!

As for the master, you did it right, it reminds me to do a patch release for 5 in our quarterly release schedule. @michaelborne

PR #439, opened 2020-02-22
Fix constructor injection with virtual inheritance

elpete:
Constructor injection with virtual inheritance doesn't work correctly if the target component does not have a constructor.  The virtual inheritance constructor is injected to the target, but none of the arguments are passed when calling the constructor.  This fixes that problem by adding the constructor arguments of the virtual inheritance mapping to the target mapping.

### DIFF
--- a/system/ioc/Builder.cfc
+++ b/system/ioc/Builder.cfc
@@ -134,32 +134,39 @@
 		var thisMap 	= arguments.mapping;
 		var oModel 		= createObject( "component", thisMap.getPath() );
 
-		// Do we have virtual inheritance?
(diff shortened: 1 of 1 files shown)


lmajano:
I need some tests for this @elpete

lmajano:
Do we have a ticket for this?

elpete:
I modified the existing tests. And no, there’s not a ticket.

lmajano:
I don't see any tests committed though?

elpete:
I’m speaking of keeping the existing virtual inheritance tests passing.

PR #440, opened 2020-03-03
Capture request before announcing onRequestCapture

elpete:
Because the `RoutingService` registered itself as an interceptor, you couldn't guarantee that it would be ran before any other `onRequestCapture` points.  This change ensures that the request is processed and the `rc` and `prc` scopes filled before calling interceptors.

### DIFF
--- a/system/web/services/RoutingService.cfc
+++ b/system/web/services/RoutingService.cfc
@@ -158,8 +158,10 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
 
 	/**
 	 * This is the route dispatcher called upon the request is captured.
+     *
+	 * @event The ColdBox Request context
 	 */
-	public void function requestCapture( event ) {
+	public void function requestCapture( required event ) {
         var rc = event.getCollection();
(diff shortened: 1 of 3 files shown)


lmajano:
@elpete did you create a companion ticket in the ColdBox Jira?

system/web/services/RoutingService.cfc
```diff
@@ -156,14 +156,13 @@ component extends="coldbox.system.web.services.BaseService" accessors="true"{
...
 
-	/****************************************************************************************************************************/
-	/* 													INTERCEPTION EVENT 														*/
-	/****************************************************************************************************************************/
-
 	/**
 	 * This is the route dispatcher called upon the request is captured.
 	 */
-	public void function onRequestCapture( event, interceptData, rc, prc, buffer ){
+	public void function requestCapture( event ) {
```
lmajano: Can you make the argument required please and add it to the cfc doc with what we are receiving please.

elpete:
https://ortussolutions.atlassian.net/browse/COLDBOX-845

PR #443, opened 2020-03-30
Updated invoke() to correctly call method by name

richardherbert:
The invoke incorrectly included "()" in with the method name which is not required


lmajano:
Thanks so much @richardherbert ! I am also impressed you added a test!  Thanks so much

richardherbert:
Always leave code in a better position than you found it :-)

PR #444, opened 2020-03-30
Change getSetting default arg to position #2

jclausen:
This will allow for the following syntax:  `getSetting( "foo", "bar" )`

This makes the syntax more intuitive to use and also matches the `getValue` method signature


lmajano:
@jclausen I think we can break compat and just switch them, have `defaultValue` first and then have the `fwSetting` last.

jclausen:
I’m OK with that.  I will update the PR with the method signature revision

lmajano:
All tests are failing @jclausen

jclausen:
I’ll dig in to it when I get back, but it looks like it’s Adobe-specific and on the Renderer tests: https://travis-ci.org/github/ColdBox/coldbox-platform/jobs/669307937?utm_medium=notification&utm_source=github_status

lmajano:
merged manually now

PR #446, opened 2020-04-02
Add option to define reinit key vs. hard-coded fwreinit

ddspringle:
Use cases have arisen for both practical and security reasons where it would be handy to be able to define the name of the variable that would allow the framework to be reinitialized. This change provides the ability to define `ReinitKey` in ColdBox.cfc config to specify that variable name, while maintaining `fwreinit` as the default if it is not defined.


lmajano:
Thanks @ddspringle I merged manually, as I had to refactor some things!  Great job!

lmajano:
https://ortussolutions.atlassian.net/browse/COLDBOX-863

PR #449, opened 2020-04-27
ColdBox-857

Tropicalista:
Fix https://ortussolutions.atlassian.net/projects/COLDBOX/issues/COLDBOX-857?filter=allopenissues&orderby=priority%20DESC

The revManifest is always in the appPath, so no need to switch to module. I'm not sure how this worked in the past.

The slash removed because already included in the path.

### DIFF
--- a/system/modules/HTMLHelper/models/HTMLHelper.cfc
+++ b/system/modules/HTMLHelper/models/HTMLHelper.cfc
@@ -2356,9 +2356,7 @@ component extends="coldbox.system.FrameworkSupertype" accessors=true singleton{
 		var currentModule 		= event.getCurrentModule();
 
 		// Calculate app path depending on module or app
-		var appPath = ( arguments.useModuleRoot && len( currentModule ) ) ?
-						controller.getSetting( "modules" ).find( currentModule ).path & "/":
-						controller.getSetting( "applicationPath" );
+		var appPath = controller.getSetting( "applicationPath" );
 		var mapping = ( arguments.useModuleRoot && len( currentModule ) ) ?
 						event.getModuleRoot() :
 						controller.getSetting( "appMapping" );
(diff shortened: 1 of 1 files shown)


lmajano:
@elpete can you validate this or is this a bug in elixir?

Tropicalista:
If you guys clarify how elixir works, I can do better implementation.

From what I understand, at the moment elixir generate js/css files always in the includes of the root. The rev-manifest is there. 

Using the arg useModuleRoot how is it at the moment, does not make much sense, because the function take in account only the name of the subpath

jclausen:
Actually, there are circumstances where a module has compiled its own assets ( and has it's own `rev-manifest` file.   This pull request would break the ability for a module to call its own manifest.

Tropicalista:
@jclausen yes, that's correct. That's why I have asked. 

Probably there must be imlemented a different way to determine if the cached path is by a module or root application. I suppose a good way is to split the cached paths to to module and root, and then search accordingly.

lmajano:
Can you help us here @jclausen

jclausen:
It seemed to me that the only problem that really needed solving by this PR was the extra slash in the path, which is also present with the module root. That can be removed for sure.

I would need clarification on what other problem needs to be solved from @Tropicalista

Tropicalista:
I try to explain better.

If I create a module with entry point "app.js" I will include it using:

```
elixirPath( fileName="js/app.js", useModuleRoot="true" )
```
I will expect that will be outputted:

```
<script src="/modules_app/MYMODULENAME/includes/js/app.js"></script>
```

At the moment the module path does not take in account the module but only the path, seraching first in the cached paths for the file name.

So the code above will just find for the filename, that's why we need to first determine what will be the base path, the root or the module.
```
// In local discovery cache?
if( variables.cachedPaths.keyExists( arguments.filename ) ){
	return variables.cachedPaths[ arguments.filename ];
}
```

lmajano:
I have now added several fixes for these issues Thanks @Tropicalista for bringing it to light!

PR #453, opened 2020-05-13
Comment out debug output in ConsoleAppender

jcberquist:
Left this debug output active, sorry 😳 . I was trying to copy the pattern in the FileAppender of what was left in place of the `out()` calls and what was commented out.

### DIFF
--- a/system/logging/appenders/ConsoleAppender.cfc
+++ b/system/logging/appenders/ConsoleAppender.cfc
@@ -152,7 +152,7 @@ component accessors="true" extends="coldbox.system.logging.AbstractAppender" {
 					var thisMessage = variables.logListener.queue[ 1 ];
 					variables.logListener.queue.deleteAt( 1 );
 
-					this.out( "writing #thisMessage.toString()#" );
+					// this.out( "writing #thisMessage.toString()#" );
 
 					if ( thisMessage.isError ) {
 						variables.error.println( thisMessage.message );
--- a/system/logging/appenders/ConsoleAppender.cfc
+++ b/system/logging/appenders/ConsoleAppender.cfc
@@ -43,8 +43,8 @@ component accessors="true" extends="coldbox.system.logging.AbstractAppender" {
 		super.init( argumentCollection = arguments );
 
 		// Output Streams
-		variables.out   = createObject( "java", "java.lang.System" ).out;
-		variables.error = createObject( "java", "java.lang.System" ).err;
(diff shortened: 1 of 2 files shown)


lmajano:
@jcberquist tests need updating. They’re failing

jcberquist:
Thanks for the nudge. I took a look, and at least locally I see two failures. The first one seems to be an easy fix, but the second one is more of an issue for me since I don't grasp the ColdBox internals. The error happens in the test of the LoaderService processShutdown() method.

```cfc
/**
 * Process the shutdown of the application
 */
LoaderService function processShutdown(){
    variables.log.info( "† Shutting down ColdBox..." );

    var wireBox = variables.controller.getWireBox();

    // Shutdown all ColdBox Scheduler Tasks, no need to delete them as WireBox will be nuked!
    variables.log.info( "† Shutting down ColdBox Task Scheduler..." );
    wirebox.getInstance( "AsyncManager@coldbox" ).shutdownAllExecutors( force = true );

    // Process services reinit
    structEach( variables.controller.getServices(), function( key, thisService ){
        variables.log.info( "† Shutting down ColdBox #arguments.key# service..." );
        thisService.onShutdown();
    } );

    // Shutdown any services like cache engine, etc.
    variables.log.info( "† Shutting down CacheBox..." );
    variables.controller.getCacheBox().shutdown();

    // Shutdown WireBox if it exists
    if ( isObject( wirebox ) ) {
        variables.log.info( "† Shutting down WireBox..." );
        wirebox.shutdown();
    }

    variables.log.info( "† ColdBox shutdown gracefully..." );

    return this;
}
```
The issue is that after all executors are shutdown via
```
wirebox.getInstance( "AsyncManager@coldbox" ).shutdownAllExecutors( force = true );
```
`variables.log.info()` continues to be called, but there is no TaskScheduler the appenders can use anymore. This is actually an issue for the FileAppender as well, it just happens that in the test harness Coldbox.cfc config, the file appender is specified as `async`, while the console appender is not. So the cfthread swallows the exception for the file appender, whereas it is evident with the console.

Like I said, I am not really sure what to do about this one.

jcberquist:
Well @lmajano not sure where to proceed from here. Locally the Lucee 5 tests are passing. Ironically, locally ACF fails on some of the async tests (in `AsyncManagerSpec.cfc`). I don't see what I would have done that would cause these to fail.

lmajano:
I will debug thanks man! Really appreciate it!

PR #454, opened 2020-05-27
Added 422 STATUS to RequestContext

TheRealAgentK:


TheRealAgentK:
Huh? I'm confused why the check failed on Lucee 5? I don't *think* this is because of my change.

lmajano:
Thanks @TheRealAgentK ! Merged. Sometimes, the lucee build fails since the latest updates , I am still trying to see why. It's always one test that is giving me trouble.

PR #456, opened 2020-08-03
COLDBOX-839 - Resolve router method/arg discrepancy

mjclemente:
Standardize to use `valuePairTranslation` instead of `valuePairTranslator`


lmajano:
@mjclemente I think I found an issue.  And yes, the discrepancy created it!  The routing services where all expecting `valuePairTranslation`.  So this change would actually introduce another bug.  So sorry about this, I should have investigated it a bit more.  Can you change it to standardize it on `valuePairTranslation` please. If you can send this to me today, I can include this in the final release.

PR #457, opened 2020-08-19
Update "Whoops" exception message to allow selecting a subset of text

homestar9:
Previously if you clicked on the exception message, it would automatically highlight all of the text which makes it difficult to select just a subset. For example, if you just needed to copy a variable name which caused the exception and not the whole message.  This change does not break the "copy to clipboard" button which will still enable users to do so if they need the entire message on their clipboard.  This change gives the feature a little more flexibility with that box.


lmajano:
Thanks, I agree with you!

PR #458, opened 2020-08-27
Correct args for setExplicitView

sjdaniels:
I believe this is the proper way to fix COLDBOX-920?

### DIFF
--- a/system/web/Renderer.cfc
+++ b/system/web/Renderer.cfc
@@ -207,7 +207,7 @@ component
 				arguments.args.append( explicitView.args, false );
 
 				// clear the explicit view now that it has been used
-				setExplicitView( "", "", {} );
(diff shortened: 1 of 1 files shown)


system/web/Renderer.cfc
```diff
@@ -207,7 +207,7 @@ component
 				arguments.args.append( explicitView.args, false );
 
 				// clear the explicit view now that it has been used
-				setExplicitView( {} );
+				setExplicitView( "", "", {} );
```
lmajano, line 210: I say, let's just remove it: `getRequestContext().removePrivateValue( "_explicitView" )`

sjdaniels:
That was my inclination too but not knowing the intricacies of the new rendering paradigm I wasn't sure. I've updated the PR accordingly.

PR #459, opened 2020-09-10
getComponentMetaData parsing error info

wpdebruin:
Add more info on getComponentMetaData errors. In lucee there's no sign of filenames or Linenumbers. I filed a bug with Lucee, but in the mean time it would be very helpful to at least have some info on which file is failing on injections.

I could make a better pull request where we rewrite most of this function to script, and add better error handling so we can pass all extended info from the original errors, so it doesn't get lost if there are filenames and numbers. ACF does this right, current versions of lucee fail to reveal the filename in getComponentMetadata parsing errors.

### DIFF
--- a/system/core/util/Util.cfc
+++ b/system/core/util/Util.cfc
@@ -219,7 +219,12 @@ Description :
 			<cfif isObject(component)>
 				<cfset md = getMetaData(component)>
 			<cfelse>
-				<cfset md = getComponentMetaData(component)>
+				<cftry>
+					<cfset md = getComponentMetadata( component )>
+					<cfcatch type="any">
+						<cfthrow message="Error parsing metaData for component #component#, please check for errors" >
+					</cfcatch>
+				</cftry>
 			</cfif>
 		</cfif>
 


system/core/util/Util.cfc
```diff
@@ -219,7 +219,12 @@ Description :
 			<cfif isObject(component)>
 				<cfset md = getMetaData(component)>
 			<cfelse>
-				<cfset md = getComponentMetaData(component)>
+				<cftry>
+					<cfset md = getComponentMetadata( component )>
+					<cfcatch type="any">
+						<cfthrow message="Error parsing metaData for component #component#, please check for errors" >
```
lmajano, line 225: @bdw429s any thoughts on this?

wpdebruin:
I talked to Zac Spitzer about this today. Actually we both looked if we could find a reproducible bug.  It is just the getComponentMetaData which is not showing any file info. (ACF has detailed info). So it makes it hard to debug until Lucee improves the error message.
In the mean time we could provide a more informative error message and keep the original info as well.
I just don't know if wrapping it in a cftry has any performance issues. If not, it is a simple change which prevents a lot of frustration :-)

lmajano:
Can we revisit this, we have changed so much.

PR #465, opened 2020-12-07
COLDBOX-952 opt for more efficient method variable merge

DominicWatson:
This change made a ~75% memory usage improvement over the previous logic. I believe that we can just do `StructAppend( ..., false )` here because the variables scope of the module is very limited. Looping over the keys in this way is basically doing the same thing but costing RAM.

Of course, only a consideration when hundreds of `renderView()` calls in a page - but that is entirely posisble with a framework.

### DIFF
--- a/system/web/RendererEncapsulator.cfm
+++ b/system/web/RendererEncapsulator.cfm
@@ -12,12 +12,7 @@
 	variables.renderedHelpers 		= {};
 
 	// Merge variables from renderer
-	variables.append(
-		attributes.rendererVariables.filter( function( key, value ){
-			return !listFindNoCase( "local,attributes,arguments", arguments.key );
-		} ),
-		true
-	);
+	structAppend( variables, attributes.rendererVariables, false );
 
 	// Localize context
 	variables.event = attributes.event;


bdw429s:
@DominicWatson if I recall, the reason for the filter is because Adobe's `variables` scope working differently than Lucee's.  Adobe stores the `local` and `arguments` scope (and possibly the `attributes`-- I'd have to check) inside of `variables`, so the filter was necessary to not accidentally overwrite unrelated variables.  So while this may not seem like a change in behavior on Lucee, it probably is on Adobe.

Can you expand on the "75% memory usage improvement"?  How did you measure this and what exactly what taking up the memory?  Your change is actual copying more, not less into the variable scope.  For complex objects however (UDFs, queries, etc), it's not actually copying at all, but simply creating references.  So I wouldn't really expect it to affect memory at all.  The only way I can conceive that the change would lower memory was if Lucee had some sort of nasty overhead for calling the filter UDF.

DominicWatson:
Re the memory, I was using FusionReactor to measure the request memory allocation. 1000 loops of renderView() with a view that had nothing in it.

The existing code was approximately 3 times slower and had ~4x the request memory allocation.

I believe that the `StructAppend( variables, attributes.rendererVariables, false )` (in particular the `false`) gets around the not overwriting things like `attributes` (which is present in Lucee) + `local` and `arguments` (which actually you overwrite manually later anyway).

I didn't see anything in the `variables` scope that you _would_ want to overwrite with whatever was in ` attributes.rendererVariables`.

As to the cause, I also tried a simple loop over the keys and building the copy that way (rather than inline {{filter}} with anon function) - had the same result. Much slower and more mem consumption than just StructAppend().

bdw429s:
@DominicWatson I've spent the past few hours testing this on Lucee 5.3.7, but I can't seem to reproduce any of your numbers.  I create a Coldbox sample app, an empty view file, and dropped this in the main view
```cfml
<cfloop times=1000>
	#renderView( 'main/empty' )#
</cfloop>
```
The existing version of the framework ran this request in around **400 ms** and FR's "Transaction Allocated Memory" was roughly **85 KB** for the request.
Making your suggested change brought the request down to around **300 ms** and **45 KB** of request memory.
So it does improve time and memory, but not on the scale you reported.  

I also tried many variations of `true` or `false` for the `overwrite` flag on the append, but it made zero difference one way or the other.  I can confirm from watching the profiler that the overhead of the closure execution is likely responsible for both the performance and memory.  If I increase the iterations to 10,000, two things come into focus
* fileExists() checks (taking about 2/3rd to 1/2 of the time)
* the structFilter and related closure calls (taking about 1/4th of the time)

The latter involves chunks like this which setup and tear down the `arguments` scope for every time the closure is called.

image

I'm fairly sure the `argument` and `local` scopes for the 1,000 UDF calls in your test likely account for the extra 40 some Kilobytes of memory.

And if I try just a basic loop over the keys in the struct
```js
	for( myVar in attributes.rendererVariables ) {
		if( !listFindNoCase( "local,attributes,arguments", myVar ) ) {
			variables[ myVar ] = attributes.rendererVariables[ myVar ];
		}
	}
```
then the request performance for the thread is basically the same as the filtering structAppend() but with memory consumption just barely higher than the non-filtering append.

My takeaway here is that anything we do is a "micro-optimization" memory wise but the most important part of it is avoiding a large number of UDF executions under load, which is quite unfortunate to see poor performance in functional coding styles in Lucee.

lmajano:
Ok, so where should we go with this optimization then? Avoid the functional style programming?

bdw429s:
My recommendation would be to switch to the `for( ... )` loop I have above to avoid the small memory overhead of the closure. However, if you really wanted to address the slowest aspect of rendering a view, cache the `fileExists()` and `expandPath()` operations that touch the file system and you'd get more overall bang for your buck than the closure removal.

lmajano:
I will update to use the for loop approach.
The fileExists() and expandPath() should only happen once and be cached in the reference maps. I am wondering which fileExists() are these.  Do you have a location for them?

lmajano:
Merged manually.

DominicWatson:
The for loop made no difference for me - it wasn't the functional style that was causing the issue - just the looping CFML generally. I still think the loop is redundant, whether you get the same performance results or not. i.e. what variables would exist that you _would_ want to be overwritten here? `StructAppend( ..., false )` avoids the issue you have with not wanting to override those variables and any other potential side effects.

Re not getting the same results, I was testing in an application with a lot of helpers, etc. - that may have made a difference re the general numbers.

But anyways, all good.

But anyways,

bdw429s:
>  I am wondering which fileExists() are these. Do you have a location for them?

There's file system calls that are defo not being cached.  To find them, dial up a test like I did above and loop render view about 10,000 times and just watch the stack trace.  You'll see them.

PR #473, opened 2021-03-04
Update BaseProxy.cfc

cichlidguru:
Added code to unloadContext to ensure any DB connections used get returned to the connection pool.

Without clearSqlProxy an executor will hold onto any connections it touched while running and they will not timeout/close, and no other code can use the connection except for the executor that last touched it.

Credit to Brad Wood for finding this!

### DIFF
--- a/system/async/proxies/BaseProxy.cfc
+++ b/system/async/proxies/BaseProxy.cfc
@@ -135,6 +135,8 @@ component accessors="true" {
 		if ( server.keyExists( "lucee" ) ) {
 			// Nothing right now
 		} else {
+			// Ensure any DB connections used get returned to the connection pool. Without clearSqlProxy an executor will hold onto any connections it touched while running and they will not timeout/close, and no other code can use the connection except for the executor that last touched it.   Credit to Brad Wood for finding this!
+			createObject( 'java', 'coldfusion.sql.DataSrcImpl' ).clearSqlProxy();
 			variables.fusionContextStatic.setCurrent( javacast( "null", "" ) );
 		}
 	}


bdw429s:
https://ortussolutions.atlassian.net/browse/COLDBOX-978

jclausen:
The only question I might have on this PR would be:  In the context of an ORM enabled application, while using ORM in thread, would the final flush of the session execute before this is called.  If not, we should wrap this with a check for an ORM enabled application and call a manual ORMFlush to make sure there is nothing left in the pool

bdw429s:
> would the final flush of the session

@jclausen So far as I know, there is no automatic flushing of any ORM sessions from inside of a thread like this.  So I would assume if you're using any ORM in a thread like this, it would be the equiv of flush at request end being off and you'd need to manually flush anything you wanted saved.

Now, we could look at how CFThread does this for example and see if there's a precedent for adding an automatic ORM flush to our asyncmanager threads.  I do think that is slightly orthogonal to this however.  If you are using ORM outside of transactions in an asyncmanager thread, i assume you already have issues with things never being flushed which also brings up an interesting point about where the Hibernate session is actually stored under the covers and if each thread in the pool has its own session (I assume it does).  Because whether or not we flush it, I doubt we're closing it either.

jclausen:
It depends on the sessions.  If `flushAtRequestEnd` is enabled then, in the context of a thread, I would think that would be the end of the thread - since a thread opens a new Hibernate session. 

If using CBORM, which requires that setting to `false`, then @lmajano would probably understand the mechanics of that.

Either way, it seems to me that addition a conditional to check for 1 ) an ORM-enabled application and 2) whether there is anything pending in the pool would be smart defensive coding

bdw429s:
> I would think that would be the end of the thread

@jclausen To be clear, there is no "end" to these threads.  They live for as long as the executor pool exists and are simply re-used.  Since they are created outside of the scope of the CF engine, none of the typical request-based lifecycle events you're thinking of would fire in this context.

> Either way, it seems to me that addition a conditional to check for 1 ) an ORM-enabled application and 2) whether there is anything pending in the pool would be smart defensive coding

I disagree on this.  It is an acceptable use case a developer to leave changes in their Hibernate session that they don't wish to Flush.  This is true of a regular HTTP request and I don't think it's the job of the asyncmanager threads to babysit that.  I do think it's worth putting some effort into ensuring any hibernate session is destroyed so it doesn't leak over to the next time that thread is used, but again that is orthogonal to the cleanup of DB connections. (i.e. it would be a different ticket)

jclausen:
> It is an acceptable use case a developer to leave changes in their Hibernate session that they don't wish to Flush. This is true of a regular HTTP request and I don't think it's the job of the asyncmanager threads to babysit that. I do think it's worth putting some effort into ensuring any hibernate session is destroyed so it doesn't leak over to the next time that thread is used, but again that is orthogonal to the cleanup of DB connections. (i.e. it would be a different ticket)

I can live with putting the onus on the flush on to the developer.  At the very least, if the pool is dirty, a simple ormClearSession() in this block would prevent any attempts to flush after this - which might throw an error in-thread.

PR #474, opened 2021-03-04
Ensures the invalid event handler is fired correctly

elpete:
Previously, if the handler existed but the action was incorrect, the invalid event lifecycle would not fire correctly.  To fix this, we override the event in the request context to be the invalid event handler event.

Fixes COLDBOX-964


bdw429s:
This pull request has been mentioned on **Ortus Solutions Community**. There might be relevant details there:

https://community.ortussolutions.com/t/coldbox-6-2-invalideventhandler-doesnt-seem-to-be-working/8789/3

PR #480, opened 2021-03-29
ACF Compatibility update for 2021 version of ORM

jclausen:
Also resolves CBORM-9 ( https://ortussolutions.atlassian.net/browse/CBORM-9 )

### DIFF
--- a/system/cache/providers/CFProvider.cfc
+++ b/system/cache/providers/CFProvider.cfc
@@ -303,7 +303,7 @@ component
 	 * @objectKey The key to retrieve
      */
     function getQuiet( required objectKey ){
-		var ACFModernEngines = [ 2018, 2021 ];
+		var ACFModernEngines = [ 2016, 2018, 2021 ];
 
 		if( ACFModernEngines.contains( server.coldfusion.productVersion.listFirst() ) ){
 			var element = getObjectStore().getQuiet( arguments.objectKey );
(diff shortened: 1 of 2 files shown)


jclausen:
Also resolves an issue where the CFProvider would never return a cache hit on ACF2021

system/cache/providers/CFProvider.cfc
```diff
@@ -303,8 +303,9 @@ component
 	 * @objectKey The key to retrieve
      */
     function getQuiet( required objectKey ){
+		var ACFModernEngines = [ 2016, 2018, 2021 ];
```
lmajano, line 306: 2016 shouldn't be added here in all reality, only 2018 plus. 
Also, why not just do a `gt 2018` for the version?

system/cache/providers/CFProvider.cfc
```diff
@@ -493,7 +494,9 @@ component
 	 * @objectKey The object cache key
 	 */
 	boolean function clearQuiet( required objectKey ){
-		if( server.coldfusion.productVersion.listFirst() == 2018 ){
+		var ACFModernEngines = [ 2016, 2018, 2021 ];
```
lmajano, line 497: 2016 shouldn't be added here in all reality, only 2018 plus.
Also, why not just do a gt 2018 for the version?

system/core/dynamic/BeanPopulator.cfc
```diff
@@ -362,7 +362,8 @@ component {
 						}
 
 						var getEntityMap = function(){
-							if ( find( "2018", server.coldfusion.productVersion ) ) {
+							var ACFModernEngines = [ 2018, 2021 ];
+							if( ACFModernEngines.contains( server.coldfusion.productVersion.listFirst() ) ){
```
lmajano: Also, why not just do a `gt 2018` for the version?

jclausen:
All tests are passing and this should be good to merge.

PR #482, opened 2021-04-07
querystring can also be a struct in relocate

wpdebruin:
Simular functionality as in buildLink, makes it easier to add params

### DIFF
--- a/system/web/Controller.cfc
+++ b/system/web/Controller.cfc
@@ -362,6 +362,15 @@ component serializable="false" accessors="true" {
 		if ( len( trim( arguments.event ) ) eq 0 ) {
 			arguments.event = getSetting( "DefaultEvent" );
 		}
+        // Query String Struct to String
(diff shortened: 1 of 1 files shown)


system/web/Controller.cfc
```diff
@@ -362,6 +362,15 @@ component serializable="false" accessors="true" {
 		if ( len( trim( arguments.event ) ) eq 0 ) {
 			arguments.event = getSetting( "DefaultEvent" );
 		}
+        // Query String Struct to String
+		if( isStruct( arguments.queryString ) ){
+			arguments.queryString = arguments.queryString
+				.reduce( function( result, key, value ){
+					result.append( "#encodeForURL( key )#=#encodeForURL( value )#" );
```
lmajano, line 369: Looks great, just do full scoping, I am getting into the habit of doing full scopes now :)

PR #483, opened 2021-04-08
Fixes issues with Adobe losing App Context in Scheduled Tasks

jclausen:
Resolves an issue where ACF would lose the app scope and usage of the original fusion context would throw a null pointer exception


lmajano:
Looks good to me @bdw429s , any thoughts on this?

jclausen:
Keep me posted on any issues and I'll jump on them.  This  is a blocker for me on some scheduled task enhancements for  project so getting this pushed out to a release is high priority.

PR #486, opened 2021-05-13
Patch/reinit fail fast adjustments

jclausen:
This PR attempts to solve several issues related to the "fail fast" approach, as well as a potentially stale/incorrect application state when an error occurs on a reinit:

1. Moves the framework readiness flag until post-afterAspectsLoad, to ensure the application, itself is fully initialized before serving traffic
2. Adds a `reinitRequestor` key to the request scope, during the reinit process, to prevent the requestor of the reinit from being trapped in the fail fast 503
3. Changes to use a full `applicationStop` to clear out potentially bad application keys if there is an error during the reinit lifecycle.  This will then fire the full `applicationStart` on next request and will prevent Coldbox from serving traffic in a partially loaded state

### DIFF
--- a/system/Bootstrap.cfc
+++ b/system/Bootstrap.cfc
@@ -156,7 +156,8 @@ component serializable="false" accessors="true" {
 						// Remove any context stragglers
 						structDelete( request, "cb_requestContext" );
 					} catch ( any e ) {
-						application.fwReinit = false;
+						// reset our application state to prevent any stale app keys from hanging around
+						applicationStop();
 						rethrow;
 					}
 				}
--- a/system/Bootstrap.cfc
+++ b/system/Bootstrap.cfc
@@ -30,8 +30,8 @@ component serializable="false" accessors="true" {
 	param name="COLDBOX_APP_KEY"       default="cbController";
 	param name="COLDBOX_APP_MAPPING"   default="";
 	param name="appHash"               default="#hash( getBaseTemplatePath() )#";
-	param name="lockTimeout" default="30" type="numeric";
-	param name="COLDBOX_FAIL_FAST" default="true";
+	param name="lockTimeout" 		   default="30" type="numeric";
+	param name="COLDBOX_FAIL_FAST" 	   default="true";
 
 	/**
 	 * Constructor, called by your Application CFC
--- a/system/Bootstrap.cfc
+++ b/system/Bootstrap.cfc
@@ -139,6 +139,7 @@ component serializable="false" accessors="true" {
 					try {
 						// Tell the word we are reiniting
 						application.fwReinit = true;
+						request.isReinitRequestor = true;
 						// Verify if we are Reiniting?
 						if (
 							structKeyExists( application, appKey ) AND application[ appKey ].getColdboxInitiated() AND needReinit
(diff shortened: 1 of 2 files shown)


bdw429s:
@jclausen Just to confirm, the `applicationStop()` is just a failsafe that won't normally fire, but kicks in if something goes south during the reinit?

jclausen:
> @jclausen Just to confirm, the `applicationStop()` is just a failsafe that won't normally fire, but kicks in if something goes south during the reinit?

Exactly, @bdw429s .  If something fails during the reinit, there's no way to tell if the framework is in a state capable of accepting live requests ( e.g. the reinit fails, for some reason, during module registration because there is a process lock on a singleton file and that DSL is now bad ).  In that case, it's better just to start with a clean slate.

jclausen:
I think it's worth mentioning that there is a couple of side effects to these changes to the loader and bootstrap:  

1. If a reinit fails, if a refresh is done, the the isFwReinit will be false on the refresh
2. If debugging `afterAspectsLoad` interceptions using an `abort` in the interception body, the `application.fwReinit` key is going to be set, but not removed - which means a reload of the page will hit the fail fast response.
  
I consider the former a necessary evil and the second the responsiblity of the developer to appropriately debug and reset for the the next attempt

system/Bootstrap.cfc
```diff
@@ -155,9 +156,9 @@ component serializable="false" accessors="true" {
 						// Remove any context stragglers
 						structDelete( request, "cb_requestContext" );
 					} catch ( any e ) {
+						// reset our application state to prevent any stale app keys from hanging around
+						applicationStop();
 						rethrow;
-					} finally {
```
lmajano, line 159: So is this the bug where ACF does NOT run any `finally` blocks?

system/Bootstrap.cfc
```diff
@@ -476,9 +477,10 @@ component serializable="false" accessors="true" {
 	boolean function onRequestStart( required targetPage ) output=true{
 		// Global flag to denote if we are in mid reinit or not.
 		cfparam( name = "application.fwReinit", default = false );
+		cfparam( name = "request.isReinitRequestor", default = false );
 
 		// Fail fast so users coming in during a reinit just get a please try again message.
-		if ( application.fwReinit ) {
+		if ( application.fwReinit && !request.isReinitRequestor ) {
```
lmajano, line 483: I am not sure how the `request.isReinitRequestor` be of use, since it only is attached to a single visitor request.  And that request is not threaded.  I am trying to see why this should be here.

system/Bootstrap.cfc
jclausen: Yes and no. With an `applicationStop` in the catch.  You wouldn't need a `finally` anyway.  That said, I've got some tweaks to this PR.  After developing for a bit with these changes, I've found that #2 in the caveat above is a pain to work with.  Also, `applicationStop`  seems to remove the Coldbox and Wirebox keys, but keeps the old FW reinit keys.  I need to do some more work with this, before we merge this change.

system/Bootstrap.cfc
jclausen: The problem is that `onRequestStart` gets called twice in a reinit, which allows a collision if a request comes in during the reinit- which then shows the "Fail Fast" message to the reinit requestor ( while the reinit continues in the logs in the background ). Effectively, the requestor gets trapped by its own reinit flag.

jclausen:
I pushed up a change to this PR to avoid the full `applicationStop` and integrated Luis' suggestions for key clearing. I'd like to keep the `request.isReinitRequestor` because it prevents the requestor from getting trapped in the fail fast.  This seems to be pretty solid - without to re-initialize the applciation scope.

system/Bootstrap.cfc
jclausen: I think we have to keep this, as insurance from the reinit requestor getting trapped in the fail fast

lmajano:
@jclausen should we continue on this path?

jclausen:
> @jclausen should we continue on this path?

I feel like we should.  The only downside to moving the framework readiness flag to after the `afterAspectsLoad` is that, if an error is thrown in that lifecycle method, your reinit flag is still up.  

The request-level variable is super handy, because it prevents the requestor from getting trapped by a fail fast message, while the reinit continues headlessly in the background.

lmajano:
Ok, tried a similar approach. Check it out pl;ease: https://github.com/ColdBox/coldbox-platform/commit/64b3dd7ae32c0a742d6e3b3a9e2d931442b9d16e

PR #487, opened 2021-05-22
CF-2018 stats.lastResult is null when task->call have runEvent("xxxx")

sanaullah:


system/async/tasks/ScheduledTask.cfc
```diff
@@ -414,19 +414,19 @@ component accessors="true" {
 
 			// After Interceptor
 			if ( isClosure( variables.afterTask ) || isCustomFunction( variables.afterTask ) ) {
-				variables.afterTask( this, variables.stats.lastResult );
+				variables.afterTask( this, (!isNull(variables.stats.lastResult)) ?:"" );
```
lmajano, line 417: I don't think you need the is null check since you are using the elvis operator my friend. Also, thinking about it let's juse use the safety navigation operator.

```
variables.afterTask( this, variables.stats?.lastResult );
```

That should safely navigate and return a result or null for us

sanaullah: Got it, will change to safe-navigation operator syntax.

lmajano:
So happy to see a pull request from you @sanaullah .  I hope you and the family are doing great!  I miss you my friend! God willing I can see you in the coming years.

PR #488, opened 2021-06-10
Proxy with app key

nockhigan:
CBORM and I would assume other proxy objects fail when using an app key. Right now the COLDBOX_APP_KEY variable in the ColdboxProxy cfc is set to cbController and can't be changed. It would probably make sense to grab the value for this from the bootstrap if it exists.

When using this with CBORM you still need to do an ormReload in your onApplicationStart function though since the cborm.models.EventHandler for the CFORM is loaded before the bootstrap is created so your app key will not be included in that component initially.


lmajano:
Thanks merged manually to test some things out.

PR #490, opened 2021-06-29
COLDBOX-1020 stop throwing an exception

Yamaha32088:


lmajano:
Can you expand on the reason behind this pull request?

Yamaha32088:
@lmajano I described it here https://ortussolutions.atlassian.net/browse/COLDBOX-1020 Basically what is happening is we have a stateless REST API built with Coldbox. Something changed in Lucee 5.3.8.189 that caused it to not set cookies the same way which causes Coldbox to throw an exception saying `Cannot find a jsessionid, URLToken or cfid/cftoken in the cookie scope. Please verify` The pull request I made allows Coldbox to proceed processing requests without any cookies being set since we have no reason to use flash storage.

These are our Application.cfc settings specific to client/session.
image

system/web/flash/ColdboxCacheFlash.cfc
```diff
@@ -59,10 +59,7 @@ component extends="coldbox.system.web.flash.AbstractFlashScope" accessors="true"
 		else if ( isSessionDefined and structKeyExists( session, "URLToken" ) ) {
 			return prefix & session.URLToken;
 		} else {
-			throw(
-				message = "Cannot find a jsessionid, URLToken or cfid/cftoken in the cookie scope. Please verify",
-				type    = "ColdboxCacheFlash.CFIDException"
-			);
+			return false;
```
lmajano, line 62: Ok, I understand the point.  I think however, relying on boolean constructs might be problematic an create extra checks that lead to performance.  After thinking of the issue, where we are stateless, I think it's best to create a unique identifier we can track for the entire request.  Because with this approach flash becomes useless.  You still might want the flash in order to last the request itself. Something like this

```
else if ( isNull( request.cbFlashId ) ){
   request.cbFlashId = createUUID();
} else {
   return request.cbFlashId;
}
```

At least this way, we still produce a key that will last for the request.

lmajano (changes requested):

system/web/flash/ColdboxCacheFlash.cfc
Yamaha32088: Updated

bdw429s:
As I just put in Slack, the correct thing to do here is to change Flash storage to use the MockFlash storage provider.

bdw429s:
@lmajano I don't think making the Flash RAM silently useless is a good solution.  Now if someone tries to use Flash storage in their app it will just silently fail which will be confusing.  A better solution would be for FlashRam to
* allow an external session context ID to be provided by the application as a valid work around for scenarios which don't use the session scope but may still have their own way of tracking repeated requests from the same client/user
* In the event that no session context ID is available, flash RAM simply goes into a disabled mode (i.e. don't bother even creating a dummy ID that serves no purpose).  Then, **if and only if** the Flash RAM is attempted to be used explicitly, throw an exception there so it will be clear when a user tries to use it why it's not working.  

This way applications who don't use Flash Ram and don't care about it don't get errors.
Applications who use flash ram but not any of the standard CF tracking IDs can still work
Applications who incorrectly try to actually use Flash RAM with no session context ID, will receive an exception _at the point they are trying to use it_.

PR #491, opened 2021-07-02
Update BeanPopulator for Hibernate 5 in Lucee

michaelborn:
The `ormGetSessionFactory().getAllClassMetadata()` syntax is deprecated (and throws an error!) in Hibernate v5+, but ColdBox only switches syntax based on the Coldfusion server name. :/ 

This PR updates ColdBox to support the newest version of the Lucee Hibernate extension.

### DIFF
--- a/system/core/dynamic/BeanPopulator.cfc
+++ b/system/core/dynamic/BeanPopulator.cfc
@@ -363,14 +363,15 @@ component {
 
 						var getEntityMap = function(){
 							try {
-								// older, deprecated syntax for pre-Hibernate5
-								return structKeyArray( ormGetSessionFactory().getAllClassMetadata() );
+								// Double array functions to convert from native java to cf java
+								// Hibernate v5+
+								return arrayToList( ormGetSessionFactory().getMetaModel().getAllEntityNames() ).listToArray();
 							} catch( any e ){
-								if ( ! listContains( e.message, "getAllClassMetadata is no longer supported" ) ){
+								if ( ! listContains( e.message, "getMetaModel" ) ){
 									rethrow;
 								}
-								// Double array functions to convert from native java to cf java
-								return arrayToList( ormGetSessionFactory().getMetaModel().getAllEntityNames() ).listToArray();
+								// Pre-Hibernate 5 syntax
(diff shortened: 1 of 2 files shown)


michaelborn:
And for anyone who asks... yes, I should write cross-engine tests for this feature, and no, I don't know how or where to start.

jclausen:
Wouldn't you want the Hibernate 5 version as the `try` and the < 5 version as the `catch`?  Since newer versions of ACF and the next versions of Lucee would be on v5.x, I would think that would be our default.

michaelborn:
That's a good point, thanks Jon! I can reverse that logic.

system/core/dynamic/BeanPopulator.cfc
```diff
@@ -362,11 +362,15 @@ component {
 						}
 
 						var getEntityMap = function(){
-							if( listFind( "2018,2021", server.coldfusion.productVersion.listFirst() ) ){
+							try {
```
lmajano, line 365: I wouldn't try the old approach first. Do the new approach first, because the old one, will only target lucee 5 with the old extension.

lmajano (changes requested):

system/core/dynamic/BeanPopulator.cfc
michaelborn: Updated. Resolved.

lmajano:
@michaelborn is there a way we can try to find out the version of hibernate instead of the try/catch. Just curious?

michaelborn:
@lmajano Yes, absolutely. I can duplicate the `getHibernateVersion()` method I just added to CBORM and add it as a ColdBox utility method instead.

https://github.com/coldbox-modules/cborm/pull/40

lmajano:
I think that maybe it's best if we abstract this out to a method in the `util.Utility` core object which is the global helper for detection.  Than we can use that here.

lmajano:
@michaelborn can you give me due date on this, i want to release 6.5 due to some Adobe ColdFusion 2021 issues

michaelborn:
@lmajano I'll work on it this morning. Expect an update today.

PR #492, opened 2021-07-02
Update internal registry on the fly as possible

timothyfarrar:
attempt to add the appender to the registry if it exists in the config but not the internal registry


lmajano:
Thanks.  Do you think you can add a test case for this please? @timothyfarrar

timothyfarrar:
Test added.

PR #493, opened 2021-07-14
Better handling of different HTTP verb methods for the same route

elpete:
Multiple methods targeting the same route but with different HTTP verbs will now correctly configure the route to call the correct action.

Additionally, actions for a specific pattern can now point to different handlers, if desired.

Resolves COLDBOX-1027


lmajano:
Wowza, why so many changes @elpete

elpete:
> Wowza, why so many changes @elpete

That wasn't my commits.  That's your [CFFormat CI commit.](https://github.com/ColdBox/coldbox-platform/pull/493/commits/253acbcb9f97225c8880e7137a098a1a2c30ce8a)

PR #494, opened 2021-07-15
Add a wirebox:targetID injection

elpete:
This is used to get the targetID (usually the mapping name) used to create an object as a property on the object.

### DIFF
--- a/system/ioc/Builder.cfc
+++ b/system/ioc/Builder.cfc
@@ -644,7 +644,11 @@ component serializable="false" accessors="true" {
 	 * @definition The dependency definition structure: name, dsl as keys
 	 * @targetObject The target object we are building the DSL dependency for
 	 */
-	private any function getWireBoxDSL( required definition, targetObject ){
+	private any function getWireBoxDSL(
+		required definition,
+		required targetID,
+		targetObject = ""
+	){
 		var thisType         = arguments.definition.dsl;
 		var thisTypeLen      = listLen( thisType, ":" );
 		var thisLocationType = "";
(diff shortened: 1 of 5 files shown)


system/ioc/Builder.cfc
```diff
@@ -644,7 +644,11 @@ component serializable="false" accessors="true" {
 	 * @definition The dependency definition structure: name, dsl as keys
 	 * @targetObject The target object we are building the DSL dependency for
 	 */
-	private any function getWireBoxDSL( required definition, targetObject ){
+	private any function getWireBoxDSL(
```
lmajano, line 647: I think this change needs to be consistent with every DSL get method, not only this one. Also, maybe this needs to update the `coldbox.system.ioc.dsl.IDSLBuilder` the `process()` method so we can standardize it.

lmajano: Also, can you expand on the ticket and here WHY do this?

lmajano (changes requested):
Can you expand more on WHY do this? Maybe add examples here of usage?

system/ioc/Builder.cfc
elpete: Sure - in Quick I have a way to get a new blank copy of the same entity.  I do this by asking WireBox for it.  But to ask WireBox for it I need to know what mapping was used.  Right now Quick does this using a `beforeInstanceAutowire`interceptor, but I thought this was cleaner.

elpete:
Hey, @lmajano.  I answered your questions earlier above.  Anything else you need for this PR?

lmajano:
> Hey, @lmajano. I answered your questions earlier above. Anything else you need for this PR?

Sorry, It was long forgotten. I am reviving it.

lmajano:
Merged manually, as I needed to add more to the interface and had to do it positionally since if not, it would be a breaking change.

PR #496, opened 2021-07-16
Announce onException interception points for async interceptors

elpete:
When trying to capture exceptions to a third-party service like
Sentry, any errors from async interceptors are lost.  By announcing
the `onException` event we can still capture these exceptions.


elpete:
This may be a good opportunity to refactor to the `AsyncManager` for this.

lmajano:
Thanks @elpete I have not refactored a lot of async code yet in ColdBox.  Basically, to give the async manager time to brew and solidify as time passes.  I think by now, we should start refactoring more code into it. Thanks for this.

PR #497, opened 2021-07-19
Fixes issue with ACF error "Routines cannot be declared more than once."

homestar9:
Addresses issue: https://ortussolutions.atlassian.net/jira/software/c/projects/COLDBOX/issues/COLDBOX-1030


lmajano:
I can't revert because this will cause issues for virtually inherited interceptors when loading the app helpers.

lmajano:
Thanks @homestar9 closed, due to the expected behavior!

PR #503, opened 2021-09-21
Added new handler metadata property `cacheIncludeRcKeys`

homestar9:
This pull request is in response to [this discussion](https://community.ortussolutions.com/t/coldbox-6-4-event-caching-and-ignoring-rc-scope/8886) in the Coldbox community.

This new feature for Coldbox gives developers more flexibility with event caching and different RC scopes. The end result should be more efficient caching and improved page load speed for many users.

**The Problem:**
Currently, event caching considers the entire `RC` scope when generating a cache key.  This means that changing any URL or FORM variable on a request will force Coldbox to generate a new cache entry. This behavior is problematic for a few reasons:
1.  A malicious attacker could take advantage and create an unlimited number of cache entries for the same page simply by adding a random URL variable. 
2. Marketing efforts often require appending a unique tracking variable to a URL (e.g. `utm_source`).  In this type of scenario, there is no need to create a new cached event for every unique URL.

**The Solution**
I was inspired by Wirebox's populator include/exclude arguments which lets developers specify a list of keys to consider when populating an entity.   My idea was to create a new metadata property called `cacheIncludeRcKeys` which would allow specifying a list of `RC` keys to include when generating a cache key.  All keys not in the list are ignored.  To maintain backward compatibility, the default value for `cacheIncludeRcKeys` is `*` which considers all keys.

How Does it Work?
Simply include a new metadata attribute  `cacheIncludeRcKeys` to a handler and specify a list of `RC` keys you want to be included in the cache key generation process.  Here are some examples:

```
function show( event, rc, prc ) cache="true" cacheTimeout="10" cacheIncludeRcKeys="slug" {
  ...
}
```

The above code will ignore all `RC` keys except for `slug` when generating a cache key.  So, if you had a route `pages/:slug` resolve to `pages.show`, the following URLs would all utilize the same cache key:
https://mysite.com/pages/foo
https://mysite.com/pages/foo?utm_source=google
https://mysite.com/pages/foo?utm_source=bing

You can also include multiple `RC` keys by specifying a list in the metadata attribute like this:
```
function show( event, rc, prc ) cache="true" cacheTimeout="10" cacheIncludeRcKeys="slug,id" {
  ...
}
```

@lmajano I know you are probably busy with preparations for Into The Box so no need to rush on this one. I'll be attending the conference in person if you would like to chat about this PR.  Looking forward to catching up with you and the entire Ortus team. :)

### DIFF
--- a/system/cache/util/EventURLFacade.cfc
+++ b/system/cache/util/EventURLFacade.cfc
@@ -35,16 +35,22 @@ component accessors="true" {
 
 	/**
 	 * Build a unique hash from an incoming request context
+     * Note: the 'event' key is always ignored from the request collection
 	 *
 	 * @event A request context object
+     * @cacheIncludeRcKeys A list of keys to include ( '*'=all keys )
 	 */
-	string function getUniqueHash( required event ){
-		var rcTarget = arguments.event
-			.getCollection()
-			.filter( function( key, value ){
-				// Remove event, not needed for hashing purposes
-				return ( key != "event" );
-			} );
+	string function getUniqueHash( required event, string cacheIncludeRcKeys="*" ){
+
+        var rcTarget = arguments.event.getCollection().filter( function( key, value ){
+			return (
+                key != "event" && // Remove event, not needed for hashing purposes
+                (
+                    cacheIncludeRcKeys == "*" || // include all keys if *
+                    listFindNoCase( cacheIncludeRcKeys, key ) // or if the key is specified
+                )
+            );
+		} );
 
 		// systemOutput( "=====> uniquehash-rcTarget: #variables.jTreeMap.init( rcTarget ).toString()#", true );
 		// systemOutput( "=====> uniquehash-rcTargetHash: #hash( variables.jTreeMap.init( rcTarget ).toString() )#", true );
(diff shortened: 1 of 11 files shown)


bdw429s:
This pull request has been mentioned on **Ortus Solutions Community**. There might be relevant details there:

https://community.ortussolutions.com/t/coldbox-6-4-event-caching-and-ignoring-rc-scope/8886/11

homestar9:
@lmajano Let me know if you have any questions or need anything else from me?  I'm curious to hear your feedback on the PR and tests.

jclausen:
This also relates to https://ortussolutions.atlassian.net/browse/COLDBOX-1203

@lmajano I would recommend merging this one in and then we can address 1203

lmajano:
@homestar9 I am so sorry we never looked at this,  if it's worth doing again, can we start again the convo on it.  I apologize, I totally missed it.

homestar9:
@lmajano thank you for following up.  Yes, my PR would add significant value for ColdBox and also boost performance.   Having the ability to use the new `cacheIncludeRcKeys` metadata attribute would be a powerful tool for developers to ensure their cached events perform optimally and can't be bypassed easily.

What are the next steps? I believe the PR should still be valid.

PR #508, opened 2021-10-31
Make unloadAll() in ModuleService more resilient

gpickin:
Make unloadAll function more resilient of broken coldbox inits, with no registeredModules key, or if the key exists with default value of an empty string.
When running tests, with the app in a subfolder, on Adobe 2018, this would stop all my tests.
Possibly showing another issue, but this makes it more resilient regardless.

This change solved all of my test failures.

### DIFF
--- a/system/web/services/ModuleService.cfc
+++ b/system/web/services/ModuleService.cfc
@@ -912,10 +912,13 @@ component extends="coldbox.system.web.services.BaseService" {
 	 * Unload all registered modules
 	 */
 	ModuleService function unloadAll(){
-		// Unload all modules
-		variables.registeredModules.each( function( key, module ){
-			unload( arguments.key );
-		} );
+		// If the registeredModules key exists, and isn't the default empty string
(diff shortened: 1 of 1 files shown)


system/web/services/ModuleService.cfc
```diff
@@ -912,10 +912,13 @@ component extends="coldbox.system.web.services.BaseService" {
 	 * Unload all registered modules
 	 */
 	ModuleService function unloadAll(){
-		// Unload all modules
-		variables.registeredModules.each( function( key, module ){
-			unload( arguments.key );
-		} );
+		// If the registeredModules key exists, and isn't the default empty string
+		if ( structKeyExists( variables, "registeredModules" ) && isStruct( variables.registeredModules ) ) {
```
lmajano, line 916: I would avoid the structkeyexists and use `isNull()` to comply with full null support.

lmajano (changes requested):

lmajano:
Merged manually, since changes were not added.

PR #509, opened 2021-11-01
Fix Elixir Path for Apps in Subfolders

gpickin:
The KEY being generated was based off the href, which included the app mapping, which is incorrect.
Since my manifest file is from the root of the app, the app mapping would never be in the key, and therefore this was not working.

By making these changes, it works as it was previously for apps in the root, but now has a chance to work in apps in subfolders.

If i ask for 
css/app.css in /gavin/publications 
it was looking for the key /gavin/publications/includes/css/app.css in the manifest - but it was /includes/css/app.css
So it usually just spits out what i put in there, 

So i was trying to hack it with /includes/css/app.css but it was just giving it back to me because /gavin/publications/includes/includes/css/app.css doesn't exist, so it works in dev, but it would never give me the manifest path... so in prod it blows up.

But with these two fixes, it uses the right thing for the key, the right thing for the href, and it adds the app mapping if you need it for outputting... so now it works.

### DIFF
--- a/system/modules/HTMLHelper/models/HTMLHelper.cfc
+++ b/system/modules/HTMLHelper/models/HTMLHelper.cfc
@@ -2583,7 +2583,7 @@ component
 		}
 		if ( mapping.len() ) {
 			variables.cachedPaths[ argumentsHash ] = "/#mapping#" & manifestDirectory[ key ];
-		} else { 
+		} else {
 			variables.cachedPaths[ argumentsHash ] = manifestDirectory[ key ];
 		}
 		return variables.cachedPaths[ argumentsHash ]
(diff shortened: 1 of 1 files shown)


elpete:
I'm not sure what the best way is to structure these packages since `coldbox-elixir` is an NPM module, not a ColdBox module, but the path functions don't make a lot of sense bundled to ColdBox.  It makes fixing bugs in the path a problem since it's tied to a ColdBox release.  It makes it hard to upgrade any conventions in Elixir since it needs to be backwards compatible with this ColdBox function as well.  I would love for the path helpers to be separate.

The only drawback I can see is that the companion ColdBox module would have just two functions and would need to be kept in lock step with the NPM module somehow.

Anyway, I think that the time of bundling elixir's path logic inside ColdBox has passed.

lmajano:
I agree with you @elpete . I think it's time for a new module and have the version in line with elixir npm.  So actually, I prefer we deprecate this feature in 6 and have the new module created @gpickin .  Do you need me to setup the new module repos?

PR #510, opened 2021-12-15
Simplify path info cleanup

bdw429s:
@lmajano Can you review this change?  I think it simplifies the two different checks you had plus it allows ColdBox to work with a file other than index.cfm.  For example, my ColdBox legacy app template uses a missing template handler to route missing files like /about.cfm to ColdBox and this code fails to remove the script name from the path info.  Note, the issue of the script name being inside the path info seems to be unique to IIS.  Other web servers don't do this.  

You had a check to remove just `/index.cfm` specifically
Then you had a check to generically remove the script name WITH a trailing slash like `/foo/something.cfm/`
And finally, you have a check to remove any remaining leading `/` slashes from the path info.

If we replace those first two checks with a generic replacement of the `cgi.script_name` then I think we'll get the same behavior, but simpler and it will work with non-index.cfm pages.

### DIFF
--- a/system/web/services/RoutingService.cfc
+++ b/system/web/services/RoutingService.cfc
@@ -1099,8 +1099,12 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
 			results[ "scriptName" ] = replaceNoCase( results[ "scriptName" ], getContextRoot(), "" );
 		}
 
-		// Clean up the from script name path_info 
-		results[ "pathInfo" ] = reReplaceNoCase( results[ "pathInfo" ], getCGIElement( "script_name", arguments.event ), "" );
+		// Clean up the from script name path_info
+		results[ "pathInfo" ] = reReplaceNoCase(
+			results[ "pathInfo" ],
+			getCGIElement( "script_name", arguments.event ),
+			""
+		);
 
 		// clean 1 or > / in front of route in some cases, scope = one by default
 		results[ "pathInfo" ] = reReplaceNoCase( results[ "pathInfo" ], "^/+", "/" );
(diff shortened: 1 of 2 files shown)


system/web/services/RoutingService.cfc
```diff
@@ -1099,17 +1099,12 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
...
-
-		// Clean the scriptname from the pathinfo if it is the first item in case this is a nested application
-		if ( len( results[ "scriptName" ] ) ) {
-			results[ "pathInfo" ] = reReplaceNoCase(
-				results[ "pathInfo" ],
-				"^#results[ "scriptName" ]#\/",
-				""
-			);
-		}
+		// Clean up the from script name path_info
```
lmajano, line 1102: Few questions.  
- Why use `reReplace` if the script name is not a regex?
- Why not use the cleaned `script_name` via `results[ "script_name" ]`? Instead of the raw value

bdw429s: @lmajano 

> Why use reReplace if the script name is not a regex?

That's your code man, lol.  The rereplace was already there.  I just changed its inputs.  You'll need to ask yourself why you used a regex-based replacement but I just kept it the way you had it.  In reality, you are wrong that's not a regex.  Look at the code-- you're making it a regex like so:
```
"^#results[ "scriptName" ]#\/",
```
by adding `^` to it so it only matches at the start of the string.  Now, that said-- this probably should be escaping any regex metacharacters which may appear in the script name as well (such as periods), but that's an existing issue with the code. 

> Why not use the cleaned script_name via results[ "script_name" ]? Instead of the raw value

Because I didn't trust it :)  `results[ "script_name" ]` is the result of some questionable find/replace above and I can't trust that to not have screwed with the value.  You're doing some sort of odd replacement of index.cfm inside the script name so I can't use that value.  I need the original, actual, real, raw value that came from the web server for this logic to work.

lmajano:
Merged manually, gracias mi hermano

PR #511, opened 2022-01-13
Fixed method return value + SQL compatibility

homestar9:
getObjectMetadata() returns a struct.  Additionally, you can't have a param "?" in MSSQL's SELECT TOP section.  I added an integer validation check to ensure arguments.max will only contain an integer.


lmajano:
Can you pull and rebase your code. All the formatting is getting in the way of me seeing the changes

homestar9:
Strange. It was such a simple change that I was able to do it via the Github website interface.  I'm about to catch a flight but will take care of it when I return.

lmajano:
merged it manually. Thanks so much kind sir!

PR #512, opened 2022-01-19
🐛 FIX: Correct trailing comma syntax error in Bootstrap.cfc

michaelborn:
This error prevents ColdBox apps from initializing on any CF engine which views trailing commas as a syntax error.


lmajano:
Merged manually tdue to all the cfformatting I had to update to the new rules.

PR #518, opened 2022-02-04
Update Response Pagination Properties for Case-Sensitive Engines

homestar9:
Depending on what engine you are using, the `Response` object's `pagination` struct will either be in camel case or all in caps.  For example in ACF, it will convert all of the variables `offset`, `maxRows`, etc... to `OFFSET`, `MAXROWS`, etc...   Since the `Response` object could be consumed by JavaScript which is case-sensitive, it's important to deliver consistent output between CFML engines.

This update addresses the problem by explicitly forcing the case of the `pagination` struct to be in camel case.

### DIFF
--- a/system/web/context/Response.cfc
+++ b/system/web/context/Response.cfc
@@ -261,11 +261,11 @@ component accessors="true" {
 		numeric totalPages   = 1
 	){
 		variables.pagination = {
-			"offset": arguments.offset,
-			"maxRows": arguments.maxRows,
-			"page": arguments.page,
-			"totalRecords": arguments.totalRecords,
-			"totalPages": arguments.totalPages
+			"offset"       : arguments.offset,
+			"maxRows"      : arguments.maxRows,
+			"page"         : arguments.page,
+			"totalRecords" : arguments.totalRecords,
+			"totalPages"   : arguments.totalPages
 		};
 		return this;
 	}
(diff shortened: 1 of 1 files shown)


homestar9:
The error in the failing Lucee test is: `Secret `SLACK_WEBHOOK` is missing. Please add it to this action for proper execution.`  I don't think it has anything to do with the PR as far as I can tell.

system/web/context/Response.cfc
```diff
@@ -260,7 +260,13 @@ component accessors="true" {
 		numeric totalRecords = 0,
 		numeric totalPages   = 1
 	){
-		variables.pagination = arguments;
+		variables.pagination = {
```
lmajano, line 263: We can't do this @homestar9 because if you do that, then the user CAN'T pass in extra arguments to include into the pagination structure.  So if we wanted to do this, then we would still have to incorporate the origination structure into the internal storage to allow for extra attributes.

lmajano: @gpickin thoughts?

homestar9: Gotcha. That makes sense.

Side note:
I've always wondered why there's a `pagination` struct in the response object anyway.  It's always smelled funny to me since it only applies when returning paginated data.  Before I was using the default Coldbox response object, I was sending paginated data as a `Page` object mementified in the `data` property like this:
```
{
	"data": {
		"pageNumber": 1,
		"maxRows": 10,
		"totalRows": 100,
		"data": [
			... // the actual paged data
		]
	},
	"error": false,
	"messages": []
}
```

lmajano:
I added another approach to get what you wanted.  Check out the latest commit.

PR #519, opened 2022-02-10
Fix for implicitViewChecks() changing global RequestContext

elpete:
When getting default layout values for implicit views, ColdBox would use the `event.setView` method in order to discover the required layout file.  This would cause problems is the `layout`/`renderLayout` call was not called from ColdBox's event lifecycle but was called manually, as is the case when using `cbmailservices.newMail().setView()` method.

This PR adds a test showing the global scope pollution.  It extracts a new `discoverLayout` method on the `RequestContext` that can be used in the `Renderer` to get the correct layout path without changing `RequestContext` values.

### DIFF
--- a/system/web/Renderer.cfc
+++ b/system/web/Renderer.cfc
@@ -577,8 +577,6 @@ component
 		boolean prePostExempt = false
 	){
 		var event                  = getRequestContext();
-		var cbox_implicitLayout    = implicitViewChecks();
-		var cbox_currentLayout     = cbox_implicitLayout;
 		var cbox_locateUDF         = variables.locateLayout;
 		var cbox_explicitModule    = false;
 		var cbox_layoutLocationKey = "";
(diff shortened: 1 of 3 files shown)


system/web/Renderer.cfc
```diff
@@ -946,7 +948,7 @@ component
 	/**
 	 * Checks if implicit views are turned on and if so, calculate view according to event.
 	 */
-	private function implicitViewChecks(){
+	private function implicitViewChecks( required string view ){
```
lmajano, line 951: I have an issue with the changes because then implicit views will be broken because nobody is setting the implicit view anymore since you change the `setView()` methods.  Did you test this?

lmajano: Also, I see the point of the pre-emptive implicit check being too pre-emptive. However, I do see this introducing a regression.

system/web/context/RequestContext.cfc
```diff
@@ -947,6 +922,39 @@ component serializable="false" accessors="true" {
 		return this;
 	}
 
+	function discoverLayout( required string view ){
```
lmajano, line 925: Please document the function

system/web/Renderer.cfc
elpete: It passed all the tests and worked in our API. If there is another test case that is missing can you add it to the test file?

lmajano:
I don't think we have a direct test for implicit views.  Just try it locally in your branch.  Create an event with no setView() and see if the view is rendered.

lmajano:
merged manually, thanks @elpete

PR #520, opened 2022-02-18
COLDBOX-1096

jclausen:
Resolves COLDBOX-1096 - SES Enabled by default through lifecycle

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,7 @@ jobs:
             ${{ runner.OS }}-commandbox-cache-${{ hashFiles( 'box.json' ) }}
 
       - name: Setup CommandBox
-        uses: Ortus-Solutions/setup-commandbox@main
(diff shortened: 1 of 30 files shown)


lmajano:
@jclausen I like where this is going. However, I think we might take it a step further.  The SES enablements where from the pre 3.x days where routing was a plugin.  At this point, there really should not be a case for turn it on/off. It will always be on as the routing service is a first class citizen now.  

I say we remove all references to any type of ses enablements.

jclausen:
@lmajano Sounds like a plan.  I am going to have to yak-shave a bunch of tests that test for non-SES URL patterns.  I will take a look at the code and update this PR.

lmajano:
Let me know when ready @jclausen

jclausen:
Thanks for the bump @lmajano . I will try to get all of the tests working tomorrow

jclausen:
@lmajano: I have updated with the recommendations.

lmajano:
Thanks @jclausen merged manually!!!

PR #521, opened 2022-02-22
Dynamically generate QueryExecute options based on config data

homestar9:
A similar solution as https://ortussolutions.atlassian.net/browse/CACHEBOX-80.
This change dynamically generates the options for each queryExecute() statement based on the configuration data so it remains consistent with the JDBCStore.


homestar9:
Not sure why the tests are failing. I don't think it is related to the change as far as I can tell.

homestar9:
@lmajano just wanted to follow up on this one in case it slipped through the cracks.  It should be safe to merge and goes along with the previous issue here: https://ortussolutions.atlassian.net/browse/CACHEBOX-80.

PR #523, opened 2022-05-09
Announce onException on async interceptor error

michaelborn:
Adds `onException` interception support for allowing error handling on async interceptions. Previously, only the `processAsyncAll()` method supported this.


lmajano:
@michaelborn do you have a ticket for this?  Great job!

PR #524, opened 2022-06-14
Resolves WIREBOX-126 and cleans up legacy conditional operators

jclausen:
This cancels out CBORM-34 and the corresponding pull request by restoring the proper fall through behavior for attribute parsing


lmajano:
Muchas gracias Señor Juan!!

PR #526, opened 2022-06-20
Updated Missing Action Response Code to 404 instead of 405

homestar9:
I believe the intent of this method is to show that a particular handler action is missing which feels more like a 404 not found error.  405 (NOT_ALLOWED) is more appropriate for invalid HTTP actions.


lmajano:
You are on a roll @homestar9 !! Thanks for all the contributions!

PR #530, opened 2022-08-02
COLDBOX-1039 - Add unlisten( closure, point ) method to InterceptorService

michaelborn:
This allows simpler removal of closure listeners:

```js
var listener = function() {
  // DO STUFF...
};
iService.listen( listener, "onCall" );

// later, when we no longer need the listener:
iService.unlisten( listener, "onCall" );
```

### DIFF
--- a/system/web/services/InterceptorService.cfc
+++ b/system/web/services/InterceptorService.cfc
@@ -264,9 +264,11 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
 	 *
 	 * @target The closure/lambda to unregister
 	 * @point  The interception point from which to unregister the listener
+	 * 
+	 * @returns True if interception point found and unregistered; else false.
 	 */
-	void function unlisten( required target, required point ){
-		unregister( "closure-#arguments.point#-#hash( target.toString() )#" );
+	boolean function unlisten( required target, required point ){
+		return unregister( "closure-#arguments.point#-#hash( arguments.target.toString() )#" );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


system/web/services/InterceptorService.cfc
```diff
@@ -259,6 +259,16 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
...
 	}
 
+	/**
+	 * Unregister the given closure from a specific interception point
+	 *
+	 * @target The closure/lambda to unregister
+	 * @point  The interception point from which to unregister the listener
+	 */
+	void function unlisten( required target, required point ){
+		unregister( "closure-#arguments.point#-#hash( target.toString() )#" );
```
lmajano, line 269: Please use `arguments.target.toString()`

system/web/services/InterceptorService.cfc
```diff
@@ -259,6 +259,16 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
 		return buffer;
 	}
 
+	/**
+	 * Unregister the given closure from a specific interception point
+	 *
+	 * @target The closure/lambda to unregister
+	 * @point  The interception point from which to unregister the listener
+	 */
+	void function unlisten( required target, required point ){
```
lmajano, line 268: Also, the `unregister` signature is a `boolean` return. So keep consistent and change it to `boolean` as well and just do a `return unregister()`

michaelborn:
Resolved feedback, @lmajano .

PR #533, opened 2022-08-10
COLDBOX-1137 - Allow passing closure as second arg to listen() and unlisten() methods

michaelborn:
I propose we add support for passing the point first and closure second. I think it's weird to pass the event name after the function:

```js
variables.interceptorService.listen( function( interceptData ){
    // do magic...
}, "ORMPreLoad" );
```
IMHO, this is more legible:

```js
variables.interceptorService.listen( "ORMPreLoad", function( interceptData ){
    // do magic...
} );
```

Resolves [COLDBOX-1137](https://ortussolutions.atlassian.net/browse/COLDBOX-1137).

### DIFF
--- a/system/web/services/InterceptorService.cfc
+++ b/system/web/services/InterceptorService.cfc
@@ -268,6 +268,7 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
 	 * @return True if interception point found and unregistered; else false.
 	 */
 	boolean function unlisten( required target, required point ){
+		arguments = normalizeListenArguments( argumentCollection = arguments );
 		return unregister( "closure-#arguments.point#-#hash( arguments.target.toString() )#" );
 	}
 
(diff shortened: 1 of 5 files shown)


system/web/services/InterceptorService.cfc
```diff
@@ -288,6 +290,23 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
...
+	/**
+	 * Allow point-first arguments when calling listen() or unlisten().
+	 *
+	 * Basically flips the `target` and `point` arguments if the former is found to be a string instead of closure.
+	 *
+	 * @target Could be the target closure... could be the listen point.
+	 * @point  Could be the interception point... could be the target closure
+	 */
+	private struct function normalizeListenArguments( required target, required point ){
+		if ( isValid( "string", arguments.target ) ) {
```
lmajano, line 302: I would probablly test if `isSimpleValue()` is faster than `isValid()`

michaelborn: You're 100% right, Luis. isSimpleValue is slightly faster in Lucee and much, much faster in ACF 2021.

https://www.trycf.com/scratch-pad/gist/c98a8a7eda138606c060883f7f9b372f

michaelborn: Done. :+1:

PR #537, opened 2022-09-06
Fix argument.exception empty/null check

homestar9:
See issue COLDBOX-1144


lmajano:
Finally figured this out. It was the exception handler and we had the same thing running 3 times.

PR #540, opened 2022-09-27
Support Event.NoRender() in Integration Tests 1151

gpickin:
https://ortussolutions.atlassian.net/browse/COLDBOX-1151

The integration test will fail if you use event.noRender() as it looks for the view file. If you return "" or relocate it works as expected.


lmajano:
Great job @gpickin very tricky indeed.

PR #541, opened 2022-09-29
🤖 TEST: Add experimental builds with the latest Lucee 5.3 beta

michaelborn:
This PR will allow us to test ColdBox against the latest Lucee RC or SNAPSHOT without halting the current ColdBox release. This will give us (and Lucee!) early warning against Lucee 5.3.x compatibility issues.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,11 +15,16 @@ jobs:
     env:
       DB_USER: root
       DB_PASSWORD: root
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         cfengine: [ "lucee@5", "adobe@2018", "adobe@2021" ]
         fullNull: [ "true", "false" ]
+        experimental: [false]
+        include:
+          - cfengine: "lucee@5.3-rc"
+            experimental: true
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2


michaelborn:
Am converting this to a draft PR until we decide on a path forward - either add a `server.json` or switch to passing `cfengine` in the `server start` command.

.github/workflows/tests.yml
```diff
@@ -15,11 +15,16 @@ jobs:
     env:
       DB_USER: root
       DB_PASSWORD: root
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         cfengine: [ "lucee@5", "adobe@2018", "adobe@2021" ]
         fullNull: [ "true", "false" ]
+        experimental: [false]
```
lmajano, line 24: Can you describe this magic :)

michaelborn: Basically, we set `experimental` to `false` by default, and `true` for the Lucee 5.3 Release Candidate. Then, if `experimental` is `true`, we set `continue-on-error` to `true` so a buggy Lucee RC won't fail the build.

michaelborn: We're doing the same to run CBORM against the Hibernate 5.4 extension without preventing CBORM releases if the Hibernate extension is buggy.

https://github.com/coldbox-modules/cborm/blob/development/.github/workflows/tests.yml#L24

lmajano: I had never seen that syntax before in github actions, do you happen to have a link to the docs so I can learn it?

lmajano:
Could you update this PR to see how it would look like if we removed all the server-*.json files?

.github/workflows/tests.yml
michaelborn: Sure. It's in the matrix documentation, under "Handling Failures".

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

michaelborn:
> Could you update this PR to see how it would look like if we removed all the server-*.json files?

My apologies @lmajano , I missed this comment! Yes, I can update to remove the `server-*.json` files.

michaelborn:
Switching back to a "Draft" PR. After discussing w/ Luis, the multiple `server-*.json` files make local development much easier, so I'm going to revert to that setup.

server.json
```diff
@@ -1,6 +1,7 @@
 {
```
lmajano, line 1: Can you remove this file entirely.  It's really not needed.

box.json
```diff
@@ -52,14 +52,13 @@
         "format":"cfformat run system/**/*.cfc,tests/specs/**/*.cfc --overwrite",
         "format:watch":"cfformat watch system/**/*.cfc,tests/specs/**/*.cfc ./.cfformat.json",
         "format:check":"cfformat check system/**/*.cfc,tests/specs/**/*.cfc",
+        "log":"server log --follow",
+        "onServerInstall":"cfpm install zip,orm,mysql,postgresql,sqlserver,document,feed,mail,debugger",
```
lmajano, line 56: this won't work, because `cfpm` looks for the server in the default location and not the overriden one we use.  Plus this will exdcute for ALL servers, and it should only do it for adobe 2021

box.json
```diff
@@ -52,14 +52,13 @@
...
         "start:lucee":"server start serverConfigFile='server-lucee@5.json' --force",
         "start:2018":"server start serverConfigFile='server-adobe@2018.json' --force",
         "start:2021":"server start serverConfigFile='server-adobe@2021.json' --force",
         "log:lucee":"server log coldbox-lucee@5 --follow",
         "log:2018":"server log coldbox-adobe@2018 --follow",
-        "log:2021":"server log coldbox-adobe@2021 --follow",
-        "cfpm":"echo '\".engine/acf2021/WEB-INF/cfusion/bin/cfpm.sh\"' | run",
```
lmajano, line 61: These are needed due to our location of the engines, you will have to restore them

box.json
```diff
@@ -52,14 +52,13 @@
...
-        "cfpm:install":"echo '\".engine/adobe2021/WEB-INF/cfusion/bin/cfpm.sh\" install ${1}' | run",
-        "install:2021":"run-script cfpm:install zip,orm,mysql,postgresql,sqlserver,document,feed,mail,debugger"
```
lmajano, line 63: This is also used by the github actions to run on adobe 2021

box.json
bdw429s: @lmajano This is actually changed now.  
* cfpm command will ignore any Lucee servers and will not error
* cfpm will search specifically for an ACF 2021+ server and ignore others
* cfpm when run as part of any server-related interception point will be smart enough to apply itself to the specific server that fired the interception.  

I had Michael change this to greatly simplify the setup.

box.json
bdw429s: @lmajano It's not necessary to call the cfpm.sh script direct.  In fact, that script won't even work on Windows!  The CommandBox cfpm command will find the script in the server home for you, and is even better because it will ensure things like the `java_home` env var are set as well.

box.json
bdw429s: @lmajano I had Michael remove these to simplify the cfpm flow.  The built in CommandBox command should be smart enough to always handle this correctly now without any special scripts.

box.json
michaelborn: My bad, looks like I accidentally dropped the `log:2021` script. I'll definitely restore that!

michaelborn: Actually, this `log:2021` script is still there - just at the bottom of the scripts.

michaelborn:
@lmajano I believe this PR is ready for a second look. Keep @bdw429s 's notes in mind when looking at the updated ACF 2021 install scripts. After all... if we REALLY needed those scripts, the build would fail, wouldnt it? :smile:

lmajano:
@michaelborn do you mind resolving the conflicts please

michaelborn:
Done @lmajano !

PR #546, opened 2022-10-27
👌 IMPROVE: Split CFFormat job into standalone PR workflow

michaelborn:
This PR fixes the broken CFFormat workflow when running on pull requests.

Using the `pull_request_target` event runs the workflow under the target repo with a read/write repo token and secret access. This is a security concern if we execute any of the untrusted code... but with careful attention this is an acceptable solution to formatting the code. ([Github suggests it in their blog on preventing PWN requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,30 @@
+name: Direct Commits
+
+on:
+  push:
+    branches-ignore:
+      - "main"
+      - "master"
+      - "development"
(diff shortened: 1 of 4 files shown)


michaelborn:
Closing this PR. Unfortunately, I do not think this or ANY approach is going to get the format-and-commit workflow to execute successfully on forked repos.

PR #550, opened 2022-11-17
🐛 FIX: Ignore lazy ORM properties

michaelborn:
Resolves lazy property evaluation errors on lazy-loaded Hibernate ORM properties, such as the following:

```js
property name="createdBy" 
            fieldtype="many-to-one" 
            cfc="User" 
            fkcolumn="FK_creationUser"
            insert="true"
            update="false"
            lazy="true"
            persistent=true;
```


michaelborn:
This PR is redundant since Luis fixed these issues in a different manner. Closing.

PR #557, opened 2023-01-24
Add a helper method to encapsulate requesttimeout to improve testing …

gpickin:
What do you think about adding a helper method to encapsulate requesttimeout to improve testing consistency.

Sometimes we're adding `setting requesttimeout=300` etc and it of course overrides the test runners timeout.
Resetting it in the beforeall() or beforeEach seems too hacky, thinking it might be a nice framework helper.

Thoughts?

### DIFF
--- a/system/FrameworkSupertype.cfc
+++ b/system/FrameworkSupertype.cfc
@@ -692,7 +692,9 @@ component serializable="false" accessors="true" {
 	 */
 	DateTimeHelper function getDateTimeHelper(){
 		if ( isNull( variables.cbDateTimeHelper ) ) {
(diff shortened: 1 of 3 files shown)


lmajano:
I am not sure what this achieves though. Can you provide examples on this/?

elpete:
The default TestBox test runner defines a timeout like so:

https://github.com/coldbox-templates/default/blob/a0610a230401e3e700f3677d6b45809ddf21cbbb/tests/runner.cfm#L1

But if we are running integration tests, our actions may override this:
```cfc
component {

    function index( event, rc, prc ) {
        setting requesttimeout="30";
        // longer runner request
    }

}
```
If our integration test calls this action, the `requesttimeout` will be overridden.

@gpickin's helper method would allow us to not do any actual changes to `requesttimeout` when using the `MockController`, avoiding this issue.

PR #558, opened 2023-01-28
Fix rev-manifest references in submodules

homestar9:
Forcing the "includes" location to the beginning of the asset path will break links to assets in sub modules. See issue for details


lmajano:
Look at ticket

PR #561, opened 2023-03-07
Feature/add module service tests

jclausen:
## Description

One thing noted in pre-release review of Coldbox 7 and its changes to module injectors, is that there are no tests present for the module service.  This PR adds a dedicated spec for the ModuleService ( which can be expanded ), and also adds a now-failing test for the classnames of models auto-mapped during module registration ( `Can register and activate an ad-hoc module` ). 

Many of our module tests check for an instance that uses the virtual mapping.  This test is failing because the class name  now uses a location from root path.

Additionally, some specs were renamed to match current casing conventions.


lmajano:
merged manually

PR #572, opened 2023-05-11
ColdBox Scheduled Tasks Updates for 6.8

GiancarloGomez:
**New Properties**
 * `annually` 
 boolean to flag task as annually scheduled
 * `delayTimeUnit`
 used to work with delays regardless of setting in chain
 * `debug`  
 used for debug output
 * `firstBusinessDay` 
 boolean to flag task as on first business day schedule
 * `lastBusinessDay` 
 boolean to flag task as on last business day schedule
 * `taskTime` 
 log of time of day for firstBusinessDay and lastBusinessDay tasks
 * `startTime`
 limits task to run on or after specific time of day
 * `endTime`
 limits task to run on or before specific time of day
 * `meta`
 struct that user can use to store any data for the task ( helpful for building UIs and not having to manage outside of task )
 * `stats.nextRun`
 missing value from docs now works 😄 

**Fixed**
  * `onFirstBusinessDayOfTheMonth()` 
  was not properly setting the value if it required to add a month
  * `onLastBusinessDayOfTheMonth()`
  was not getting the correct value
  * `delay()` 
  would overwrite the time unit based when it was defined in the chain, now it gets checked in `start()` and only applied if the `timeUnit` matches the `delayTimeUnit` or if the `timeUnit` is **"seconds"**, which will convert the delay value to seconds from the `delayTimeUnit` setting.
  * `isConstrained()` 
    * **updated check for dayOfTheMonth**
    If the day the user intended to run is higher than the last day of the current month then we allow it to run on the last day of the month.
     * Added checks for new `startTime` and `endTime` properties

**Simplified Functions**
 * `everyDay()`
 * `everyWeek()`
 * `everyMonth()`

**Updated Functions**
* Updated `run()` with force argument, to allow to run the task on demand, even if it is paused.

**New Functions**
  * `debug()` 
  used for setting debug setting ( can also be done in task init )
  * `startOnTime()` 
  used to set variables.startTime
  * `endOnTime()`
  used to set variables.endTime
  * `between()`
  used to set variables.startTime and variables.endTime in one call
  * `setMeta()`
  used to set variables.meta
  * `setMetaKey()`
  used to add / save a key to variables.meta
  * `deleteMetaKey()`
  used to delete a key from variables.meta

**New Private Functions**
 * `getLastBusinessDayOfTheMonth()`
 * `getFirstBusinessDayOfTheMonth()`
 * `setInitialNextRunTime()`
 * `setNextRunTime()`
 * `debugLog()`

**Known Bugs**
The fix for delay is temporary as we need to support other timeUnit conversions and update the documentation.

## Jira Issues
COLDBOX-1226

## Type of change
- [x] Improvement

### DIFF
--- a/system/async/tasks/ScheduledTask.cfc
+++ b/system/async/tasks/ScheduledTask.cfc
@@ -506,9 +506,13 @@ component accessors="true" {
 	 * - when
 	 * - dayOfTheMonth
 	 * - dayOfTheWeek
+	 * - firstBusinessDay
 	 * - lastBusinessDay
-	 * - weekends
 	 * - weekdays
+	 * - weekends
+	 * - startOnDateTime
+	 * - endOnDateTime
+	 * - startTime and/or endTime
 	 *
 	 * This method is called by the `run()` method at runtime to determine if the task can be ran at that point in time
 	 */
@@ -537,6 +541,14 @@ component accessors="true" {
 			return true;
 		}
 
+		// Do we have day of the week?
+		if (
+			variables.dayOfTheWeek > 0 &&
+			now.getDayOfWeek().getValue() != variables.dayOfTheWeek
+		) {
+			return true;
+		}
+
 		// Do we have a first business day constraint
 		if (
 			variables.firstBusinessDay &&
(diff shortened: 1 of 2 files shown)


GiancarloGomez:
Did not create separate branch off release branch - posting new PR in a few

PR #585, opened 2023-09-30
Add Ability to Exclude LogBox Appenders to Root Logger and Categories

homestar9:
This PR adds the ability to define LogBox appender exclusions to the root logger and categories by using a special `exclude` key similar to Wirebox's object populator.

Tests included, of course. :)

Addresses: https://ortussolutions.atlassian.net/browse/LOGBOX-70


homestar9:
@lmajano not sure what is up with the Lucee 6 tests, but want to make sure you saw this since I submitted it before Summit.

PR #589, opened 2024-03-12
Add exception for set-cookie headers to always append

homestar9:
# Description

See https://ortussolutions.atlassian.net/browse/COLDBOX-1269 for a discussion on why this PR was created.

Important: Before merging, check line 1816. I'm unclear on the purpose of the `responseHeaders` value in the `variables` scope, but that value will always be overwritten.  If you want to append the header there too, let me know and I will amend the PR.

## Type of change

Please delete options that are not relevant.

- [ ] Bug Fix
- [ ] Improvement


lmajano:
That's fine, that one is for internal trqacking

PR #595, opened 2024-12-10
Use a callback to handle dynamic getHttpHeader calls in testing

gpickin:
Here is a POC to fix tests which fail due to headers not being dynamic. 
This should solve the cbotel errors because it expects headers, or default values to work. 

This has been painful for a while, especially in more dynamic testing flows.

I'll let you decide if its worth while or not to make a ticket etc.


lmajano:
This looks like a good addition, can you create a ticket, and a test for it, please

lmajano:
I still don't have a ticket or a test @gpickin

PR #607, opened 2025-05-30
[COLDBOX-1331] Add ability to ignore or include RC keys when caching events

homestar9:
@lmajano 
This pull request is in response to [this discussion](https://community.ortussolutions.com/t/coldbox-6-4-event-caching-and-ignoring-rc-scope/8886) in the Coldbox community and is a revival of the original PR here: https://github.com/ColdBox/coldbox-platform/pull/503

This new feature for Coldbox gives developers more flexibility with event caching and different RC scopes. The end result should be more efficient caching and improved page load speed for many users.

**The Problem:**
Currently, event caching considers the entire RC scope when generating a cache key. This means that changing any URL or FORM variable on a request will force Coldbox to generate a new cache entry. This behavior is problematic for a few reasons:

A malicious attacker could take advantage and create an unlimited number of cache entries for the same page simply by adding a random URL variable.
Marketing efforts often require appending a unique tracking variable to a URL (e.g. utm_source). In this type of scenario, there is no need to create a new cached event for every unique URL.

**The Solution**
I was inspired by Wirebox's populator include/exclude arguments which lets developers specify a list of keys to consider when populating an entity. My idea was to create a new metadata property called cacheIncludeRcKeys which would allow specifying a list of RC keys to include when generating a cache key. All keys not in the list are ignored. To maintain backward compatibility, the default value for cacheIncludeRcKeys is * which considers all keys.

**How Does it Work?**
Simply include a new metadata attribute cacheIncludeRcKeys to a handler and specify a list of RC keys you want to be included in the cache key generation process. Here are some examples:

```
function show( event, rc, prc ) cache="true" cacheTimeout="10" cacheIncludeRcKeys="slug" {
  ...
}
```
The above code will ignore all RC keys except for slug when generating a cache key. So, if you had a route pages/:slug resolve to pages.show, the following URLs would all utilize the same cache key:
https://mysite.com/pages/foo
https://mysite.com/pages/foo?utm_source=google
https://mysite.com/pages/foo?utm_source=bing

You can also include multiple RC keys by specifying a list in the metadata attribute like this:

```
function show( event, rc, prc ) cache="true" cacheTimeout="10" cacheIncludeRcKeys="slug,id" {
  ...
}
```

### DIFF
--- a/system/cache/util/EventURLFacade.cfc
+++ b/system/cache/util/EventURLFacade.cfc
@@ -73,7 +73,7 @@ component accessors="true" {
         } 
 
         // Cache Excludes
-        if ( len( arguments.eventDictionary?.cacheExclude ) ) {
+        if ( len( arguments.eventDictionary.cacheExclude ) ) {
             // Blacklist specific keys
             var excludeKeys = arguments.eventDictionary.cacheExclude.listToArray();
             rcTarget = rcTarget.filter( ( key, value ) => {
--- a/system/cache/util/EventURLFacade.cfc
+++ b/system/cache/util/EventURLFacade.cfc
@@ -57,15 +57,12 @@ component accessors="true" {
         // 2. Include specific keys (if `cacheInclude` is not "*")
         // 3. Exclude specific keys (if `cacheExclude` is provided and not empty)
         
-        // Use custom filter closure stored in dictionary
-        if ( len( arguments.eventDictionary?.cacheFilter ) ) {
-			rcTarget = arguments.event.getController()
-				.runEvent(
-					event = arguments.event.getCurrentHandler() & "." & arguments.eventDictionary.cacheFilter,
-					eventArguments = { rcTarget : rcTarget },
-					private   = true,
-					requestContext = arguments.event
-				);
+        // Use custom filter closure/UDF stored in dictionary
+        if ( 
+            isClosure( arguments.eventDictionary.cacheFilter ) || 
+            isCustomFunction( arguments.eventDictionary.cacheFilter )
+        ) {
+			rcTarget = arguments.eventDictionary.cacheFilter( rcTarget );
         } 
         
         // Cache Includes
--- a/system/cache/util/EventURLFacade.cfc
+++ b/system/cache/util/EventURLFacade.cfc
@@ -38,39 +38,69 @@ component accessors="true" {
      * Note: the 'event' key is always ignored from the request collection
 	 *
 	 * @event A request context object
-     * @cacheIncludeRcKeys A list of keys to include ( '*'=all keys )
+     * @targetContext The targeted request context object
+     * @eventDictionary The event metadata containing cache annotations
 	 */
-	string function getUniqueHash( required event, string cacheIncludeRcKeys="*" ){
+	string function getUniqueHash( 
+        required event,
+        required struct eventDictionary
+    ){
+        
+        // Assign the RC struct and filter out the "event" key, which is not needed for cache keys
+        var rcTarget = arguments.event.getCollection().filter( ( key, value ) => { 
+            return key != "event";
+        } );
+        
+        // Apply cache key filtering based on annotations
+        // We apply them in the following order:
+        // 1. Custom filter closure (if provided)
+        // 2. Include specific keys (if `cacheInclude` is not "*")
+        // 3. Exclude specific keys (if `cacheExclude` is provided and not empty)
+        
+        // Use custom filter closure stored in dictionary
+        if ( len( arguments.eventDictionary?.cacheFilter ) ) {
+			rcTarget = arguments.event.getController()
+				.runEvent(
+					event = arguments.event.getCurrentHandler() & "." & arguments.eventDictionary.cacheFilter,
+					eventArguments = { rcTarget : rcTarget },
+					private   = true,
+					requestContext = arguments.event
+				);
+        } 
+        
+        // Cache Includes
+        // only process if cacheInclude isn't set to "*"
+        if ( !arguments.eventDictionary.cacheInclude == "*" ) {
+            // Whitelist specific keys
+            var includeKeys = arguments.eventDictionary.cacheInclude.listToArray();
+            rcTarget = rcTarget.filter( ( key, value ) => { 
+                return includeKeys.findNoCase( key ) > 0;
+            });
+        } 
 
-        var rcTarget = arguments.event.getCollection().filter( function( key, value ){
-			return (
-                key != "event" && // Remove event, not needed for hashing purposes
-                (
-                    cacheIncludeRcKeys == "*" || // include all keys if *
-                    listFindNoCase( cacheIncludeRcKeys, key ) // or if the key is specified
-                )
-            );
-		} );
+        // Cache Excludes
+        if ( len( arguments.eventDictionary?.cacheExclude ) ) {
+            // Blacklist specific keys
+            var excludeKeys = arguments.eventDictionary.cacheExclude.listToArray();
+            rcTarget = rcTarget.filter( ( key, value ) => {
+                return excludeKeys.findNoCase( key ) == 0;
+            });
+        }
 
-		// systemOutput( "=====> uniquehash-rcTarget: #variables.jTreeMap.init( rcTarget ).toString()#", true );
-		// systemOutput( "=====> uniquehash-rcTargetHash: #hash( variables.jTreeMap.init( rcTarget ).toString() )#", true );
+        var targetMixer = {
+            // Get the original incoming context hash
+            "incomingHash" : hash( variables.jTreeMap.init( rcTarget ).toString() ),
+            // Multi-Host support
+            "cgihost"      : buildAppLink()
+        };
 
-		var targetMixer = {
-			// Get the original incoming context hash
-			"incomingHash" : hash( variables.jTreeMap.init( rcTarget ).toString() ),
-			// Multi-Host support
-			"cgihost"      : buildAppLink()
-		};
+        // Incorporate Routed Structs
+        targetMixer.append( arguments.event.getRoutedStruct(), true );
 
-		// Incorporate Routed Structs
-		structAppend(
-			targetMixer,
-			arguments.event.getRoutedStruct(),
-			true
-		);
+        
 
-		// Return unique identifier
-		return hash( targetmixer.toString() );
+        // Return unique identifier
+        return hash( targetmixer.toString() );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


homestar9:
@lmajano I see some tests failed, but I do not think they are related to my PR.

homestar9:
Hi @lmajano, I just wanted to ensure you saw this PR you requested me to create again.  This would be a really powerful addition to Coldbox and help many users who use event caching, but also rely on marketing parameters like UTM variables, which can pollute the RC scope and bypass event caching.

jclausen:
I think this is a really good addition.  The name `cacheIncludeRcKeys` is a little awkward,  but until we nest the `event` config, a more descriptive name is probably better.  I might suggest `eventCacheInclude` and then support a syntax like.  `eventCacheInclude = "rc.*,prc.authUserId"` ( defaulted to `rc.*` ).   You wouldn't have to use `isDefined`.  You could just check known contexts, split off the prefix and then filter on those keys.  A bit more code, but more powerful. 

The reason for the abstraction is that, authentication info might be retained in the `prc` rather than exposed in the `rc`, and it would be handy to be able to handle granular caching needs that are behind authentication.

In the function arguments, though, I think you can use shorter argument names, because the context is already clear. 

I would also suggest adding handling for this at the function annotation level:

```
function myAction( event, rc, prc ) eventCacheInclude="rc.*,prc.authUserId"{
...
}
```

homestar9:
@jclausen can you think of a faster way to isolate the `rc` and `prc` keys? Performance is paramount, as all this will occur on every request.

```
// Build two arrays of keys to include: rcKeys and prcKeys
var rcKeys  = [];
var prcKeys = [];
for ( var a=1; a <= listLen(cacheInclude); a++ ) {
	var part = trim( listGetAt( cacheInclude, a ) );
	var dots = listToArray( part,"." );
	if ( dots.len() == 2 ) {
		var scope = dots[ 1 ];
		var key   = dots[ 2 ];
		if ( scope == "rc" ) {
			arrayAppend( rcKeys, key );
		}
		else if ( scope == "prc" ) {
			arrayAppend( prcKeys, key );
		}
	}
}
```

homestar9:
@lmajano This update adds support for more granular control over cache key generation via event metadata. Based on your feedback, the following enhancements are now supported in the `EventURLFacade`:

`cacheFilter`: When specified, the framework will look for a private method in the handler matching the given name. This method receives a struct, `rcTarget`, containing request collection data, which can be modified as needed before the cache key is generated.

`cacheInclude` (defaults to *): Limits the cache key to only the specified keys in the request collection.

`cacheExclude`: Excludes the specified keys from being used in the cache key.

These filters are not mutually exclusive, giving developers flexibility to use one or more depending on their use case. However, the order is important. We execute them in this order: `cacheFilter` --> `cacheInclude` --> `cacheExclude`

Test coverage is included to validate different combinations and behaviors. Please let me know if you have any feedback or run into any issues.

tests/specs/integration/EventCachingSpec.cfc
```diff
@@ -24,13 +24,354 @@
...
+                // execute an event and specify a queryString variable
+                // in this test we know that the cacheFilter will randomize the slug and id keys
+                var event1 = execute( 
+                    event = "eventcaching.withAllFilters", 
+                    renderResults = true,
+                    queryString="id=1&slug=foo&utm_source=google" 
+                );
+				var prc1 = event1.getPrivateCollection();
+
+                expect( prc1.cbox_eventCacheableEntry ).toBeStruct().toHaveKey( "cacheFilter,cacheInclude,cacheExclude" );
```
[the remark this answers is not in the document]
homestar9: Commas are a valid way to ensure the struct has all keys.

system/web/services/RequestService.cfc
```diff
@@ -153,12 +153,12 @@ component extends="coldbox.system.web.services.BaseService" {
 			}
 
 			// Incorporate metadata about event
-			eventCache.append( eventDictionary, true );
+        eventCache.append( eventDictionary, true );
```
lmajano, line 156: formatting please

homestar9: Fixed indentation

system/cache/util/EventURLFacade.cfc
```diff
@@ -35,36 +35,72 @@ component accessors="true" {
...
+        } );
+        
+        // Apply cache key filtering based on annotations
+        // We apply them in the following order:
+        // 1. Custom filter closure (if provided)
+        // 2. Include specific keys (if `cacheInclude` is not "*")
+        // 3. Exclude specific keys (if `cacheExclude` is provided and not empty)
+        
+        // Use custom filter closure stored in dictionary
+        if ( len( arguments.eventDictionary?.cacheFilter ) ) {
```
lmajano, line 61: Also note that the ?. is not necessary anymore, since we are seeding all metadata with it.
So the valuation is `!isSimpleValue( )`

system/cache/util/EventURLFacade.cfc
```diff
@@ -35,36 +35,72 @@ component accessors="true" {
...
+            return key != "event";
+        } );
+        
+        // Apply cache key filtering based on annotations
+        // We apply them in the following order:
+        // 1. Custom filter closure (if provided)
+        // 2. Include specific keys (if `cacheInclude` is not "*")
+        // 3. Exclude specific keys (if `cacheExclude` is provided and not empty)
+        
+        // Use custom filter closure stored in dictionary
```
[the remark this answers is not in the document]
homestar9: I had an idea to accomplish what I think you're looking for... what do you think of this theoretical solution?

```
// HandlerService.cfc
private struct function getEventCachingMetadata( required ehBean, required oEventHandler ){
    // ...
    
    mdEntry.cacheFilter = arguments.ehBean.getActionMetadata( "cacheFilter", "" );

    // if the cacheFilter is a closure, then we store it in the mdEntry
    if ( 
        len( mdEntry.cacheFilter ) && 
        (
            isClosure( arguments.oEventHandler[ mdEntry.cacheFilter ] ) || 
            isCustomFunction( arguments.oEventHandler[ mdEntry.cacheFilter ] )
		)
    ) {
        mdEntry.cacheFilter = arguments.oEventHandler[ mdEntry.cacheFilter ];
    }
    
    // ...
```
Then...

```
// EventUrlFacade.cfc
string function getUniqueHash( 
	required event,
	required struct eventDictionary
){

	// ...
	
	// Use custom filter closure stored in dictionary
	if ( isClosure( arguments.eventDictionary.cacheFilter ) ) {
		rcTarget = arguments.eventDictionary.cacheFilter( rcTarget );
	} 
	
	// ...
```

The only catch is that the custom function/closure can't be private. Otherwise, it's not accessible in `oEventHandler`

system/cache/util/EventURLFacade.cfc
homestar9: I think my latest PR addresses this.

homestar9:
@lmajano, when you have a minute, could you please check again to see if it meets your expectations? I am currently working on updating the documentation.

system/web/services/HandlerService.cfc
```diff
@@ -689,17 +692,47 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
...
+							isClosure( arguments.oEventHandler.EVENT_CACHE_SUFFIX ) || 
+                            isCustomFunction( arguments.oEventHandler.EVENT_CACHE_SUFFIX )
 						) {
 							mdEntry.suffix = oEventHandler.EVENT_CACHE_SUFFIX( arguments.ehBean );
 						} else {
 							mdEntry.suffix = arguments.oEventHandler.EVENT_CACHE_SUFFIX;
 						}
+
+                        // if the cacheFilter has a length and is a method, then we need to verify and store the resulting closure
+                        if ( len( mdEntry.cacheFilter ) ) {
```
lmajano, line 710: This won't work on a closure/lambda.  Just do if `!isSimpleValue( mdEntry.cacheFilter )`

system/web/services/HandlerService.cfc
```diff
@@ -689,17 +692,47 @@ component extends="coldbox.system.web.services.BaseService" accessors="true" {
...
+                            // if the method doesn't exist, then throw an exception
+                            if ( !arguments.oEventHandler._actionExists( mdEntry.cacheFilter ) ) {
+                                throw( 
+                                    message = "CacheFilter expected a private method '#mdEntry.cacheFilter#'",
+                                    type = "HandlerInvalidCacheFilterException", 
+                                    detail = "CacheFilter method '#mdEntry.cacheFilter#' does not exist in handler '#getMetaData( oEventHandler ).name#'. Please verify your cacheFilter annotation."
+                                );
+                            }
+
+                            mdEntry.cacheFilter = arguments.oEventHandler._privateInvoker( mdEntry.cacheFilter, {} );
```
lmajano, line 721: This looks good

lmajano:
Fantastic job. Few moer changes and this is solid gold!

system/cache/util/EventURLFacade.cfc
```diff
@@ -35,36 +35,66 @@ component accessors="true" {
...
+        if ( arguments.eventDictionary.cacheInclude != "*" ) {
+            // Whitelist specific keys
+            var includeKeys = arguments.eventDictionary.cacheInclude.listToArray();
+            rcTarget = rcTarget.filter( ( key, value ) => { 
+                return includeKeys.findNoCase( key ) > 0;
+            });
+        } 
+
+        // Cache Excludes
+        if ( len( arguments.eventDictionary?.cacheExclude ) ) {
```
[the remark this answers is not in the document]
homestar9: Should be fixed in the latest commit

lmajano:
What a solid PR @homestar9 ! Thanks so much for this, what a great addition. Do you mind sending me a pr for the docs now?

homestar9:
Thank you so much for the positive feedback! I'm almost done with the docs, but I accidentally wrote them on the wrong Gitbook branch (ugh!). I'll get that fixed up and let you know as soon as it's completed.

homestar9:
@lmajano, I'm having trouble finding the instructions on how to create multiple change request branches on Gitbook (ugh!). However, if you look at change request #72, you should see the updates needed for this PR as well as for an older PR related to module configuration, which was never approved. See below:

image

The green box is for this PR.  The yellow box is for an older PR.

lmajano:
Hopefully this week. I have a few issues to address, before I can do a 8.0 release.

lmajano:
Where you able to send me a pr for the docs?

homestar9:
@lmajano I thought I did.  I see it in Gitbook with a status of "In review"

image

PR #609, opened 2025-06-15
Rename 'cfsqltype' to 'sqltype'

michaelborn:
To pass a query param sql type _without `bx-compat-cfml` installed_, you _must_ use the `sqltype` notation. (Not `cfsqltype`.

This fixes the following error in the DB appender and JDBC store, caused by the query param not being recognized as a `datetime` type:

```
Caused by: com.mysql.cj.jdbc.exceptions.MysqlDataTruncation: Data truncation: Incorrect datetime value: '{ts '2025-06-10 19:35:29'}' for column 'logdate' at row 1
...
```


lmajano:
Thanks, I will have to do this manually, in order to add compat.

PR #631, opened 2025-11-26
Fix for COLDBOX-1348

GunnarLieb:
Problem

 After upgrading to ColdBox 7.5.2, the Renderer.cfc was calling discoverViewPaths() to locate a layout before checking if the layout name was
  empty. When a module had layoutParentLookup = true and called event.noLayout() (which sets the layout to an empty string), the framework tried to
  locate a layout with an empty name, causing the error:

  The layout [] was not found in the module path: [/modules_app/akibase/layouts/]

  Solution

  I reorganized the code in /coldbox/system/web/Renderer.cfc:608-637 to:
  1. First check if the layout is empty (len( cbox_currentLayout ) eq 0)
  2. If empty, just render the view without any layout discovery
  3. Only call discoverViewPaths() when we know the layout is not empty

### DIFF
--- a/system/web/Renderer.cfc
+++ b/system/web/Renderer.cfc
@@ -605,19 +605,23 @@ component
 			}
 		}
 
-		// Discover the layout location + helpers
-		var layoutLocations = discoverViewPaths(
-			view          : cbox_currentLayout,
-			module        : arguments.module,
-			explicitModule: cbox_explicitModule,
-			isLayout      : true
-		);
-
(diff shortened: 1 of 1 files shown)


lmajano:
@GunnarLieb I totally dropped the ball on this.  Can you reopen this but on v8?

lmajano:
We have completed;y changed resolution now to account for this and more. Please test in be

PR #678, opened 2026-08-15
feat: HTTP caching primitives - ETag, Last-Modified, Cache-Control

lmajano:
# Description

Implements the Tier 1 (Event Caching-integrated) HTTP caching primitives from `docs/specs/http-caching.md`, plus the standalone manual API:

- **`RequestContext`**: `event.etag()` / `event.lastModified()` / `event.cacheControl()`, a new `isNoExecution()` predicate, and a portable `toHTTPDate()` formatter (built from individual date parts rather than a `dateTimeFormat()` mask, since CFML's classic mask letters and Java's `DateTimeFormatter` pattern letters are different dialects and not safe to assume across BoxLang/Lucee/Adobe). `etag()`/`lastModified()` implement the full RFC 7232 conditional-GET matching rules: `If-None-Match` may be `*` or a comma-separated list and is always compared weakly, and a request carrying `If-None-Match` ignores `If-Modified-Since` entirely per §3.3.
- **`Response`**: matching `withETag()` / `withCacheControl()` fluent helpers.
- **`HandlerService`**: new `etag` / `etagWeak` / `lastModified` / `cacheControl` annotations alongside the existing `cache` / `cacheTimeout` event-caching annotations. No separate on/off switch - these annotations are read unconditionally, gated only by the existing `cache="true"` on the action, exactly like `cacheInclude`/`cacheExclude`/`cacheFilter` already are.
- **`Bootstrap`**: on a cache write, computes an ETag hash and/or Last-Modified timestamp once and stores it (plus the `etagWeak` flag) on the cache entry; on a cache hit, replays them through the same `event.etag()`/`event.lastModified()` matching logic as the manual API, rather than a separate narrower implementation - a match skips the body replay entirely and sends a bare `304`. A handler that never sets the new annotations sees no behavior change.
- **`RestHandler`**: `aroundHandler` now also guards on `isNoExecution()`, so a conditional-GET resolved inside an action doesn't hit the same write-after-commit hazard the SSE `isSSE()` guard already covers.

Two real bugs were found and fixed while writing tests against actual execution rather than trusting the code: `cacheControl()`'s `isBoolean(val)` check is loosely true for any castable value (`isBoolean(60)` is `true` in CFML/BoxLang), which silently dropped numeric directive values like `max-age=60` down to a bare `"max-age"` token; and the original `dateTimeFormat()` mask threw `"Too many pattern letters: d"` on BoxLang, which is what led to the portable `toHTTPDate()` implementation.

A Copilot review pass on this PR additionally found: `If-None-Match` matching was exact-string-only (missing `*`/lists/weak comparison); `If-Modified-Since` wasn't ignored when `If-None-Match` was present; `toHTTPDate()` labeled its output "GMT" without actually converting to UTC first; the cache-hit replay path duplicated (and under-implemented) the matching logic instead of reusing `etag()`/`lastModified()`; and the `etagWeak` flag was computed at write time but never persisted onto the cache entry. All fixed, with new test coverage for each.

## Testing

Unit coverage for `etag()`/`lastModified()`/`cacheControl()`/`isNoExecution()`/`toHTTPDate()` (engine-agnostic, no BoxLang gate — pure header logic), `Response` fluent header tests, `RestHandler.aroundHandler` guard tests, and integration tests extending the existing `EventCachingSpec` with two new test-harness handler actions. The integration suite could not be executed in the local sandbox used for this work (`BaseIntegrationTest` needs a real servlet CGI scope unavailable there — confirmed this is a pre-existing limitation, not a regression, by running the unmodified spec and observing the identical failure), but the new handler actions and helper methods were verified directly via bare instantiation outside the full ColdBox bootstrap. Full local regression suite: 200 passed, 4 failed/23 errors — unchanged against the established pre-existing baseline (21 errors) plus 2 pre-existing `RestHandlerTest` sandbox-limitation errors.

## Jira Issues

[COLDBOX-1415](https://ortussolutions.atlassian.net/browse/COLDBOX-1415)

## Type of change

- [ ] Bug Fix
- [ ] Improvement
- [x] New Feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

(The spec doc in this same PR is the documentation update.)

## Checklist

- [x] I have made corresponding changes to the documentation
- [x] New and existing unit tests pass locally with my changes

[COLDBOX-1415]: https://ortussolutions.atlassian.net/browse/COLDBOX-1415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/specs/http-caching.md
+++ b/docs/specs/http-caching.md
@@ -1,0 +1,587 @@
+# Spec: HTTP Caching Primitives in ColdBox
+
+**Status:** Draft — no implementation yet
+**Target:** ColdBox 8.3.0 (or next minor)
+**Runtime:** BoxLang + CFML (Adobe, Lucee) — pure HTTP header mechanics, no BIF dependency
+**Related:** ColdBox's existing Event Caching (`system/Bootstrap.cfc`, `HandlerService.cfc`);
+`docs/specs/sse-streaming.md` (the annotation/interception-point conventions this spec follows)
+
+---
+
+## 1. Motivation
+
+A case-insensitive grep across `system/` for `etag`, `last-modified`, `cache-control`,
+`if-none-match`, `if-modified-since`, and `304` returns **zero hits** (the lone `"304"` string
+anywhere in the codebase is a status-text lookup entry, unrelated to caching). ColdBox has no
+concept of HTTP-level conditional requests or cache negotiation. Every response — cached
+server-side or not — always sends a full `200` with a full body.
+
+That is a real gap for anything that talks to a browser, CDN, or reverse proxy: API resources
+that rarely change, static-ish content endpoints, polling clients, HTMX partials. All of them
+would benefit from being able to say "you already have this" (`Cache-Control`) or "ask me, and
+I'll say 304 if nothing changed" (`ETag` / `Last-Modified`).
+
+### This is not Event Caching, and must not be confused with it
+
+ColdBox already ships a caching system that looks adjacent but solves a different problem, and
+this spec's central design move is to sit *on top of* it rather than duplicate it.
+
+**Event Caching** (`cache="true"` on a handler action, `system/Bootstrap.cfc:243-403`) caches the
+**server's own rendered output** in CacheBox so ColdBox can skip re-running the handler and
+re-rendering the view on a hit. It is a *server compute* optimization. It has nothing to say to
+the client — a cache hit still ships a full `200` response with the full body over the wire,
+every time, forever, to every client, even one that already has last minute's identical bytes.
+
+**HTTP Caching** (this spec) is the *client-facing* layer: it lets the browser, a CDN, or a
+reverse proxy skip the round trip entirely, or lets the server skip sending the body when the
+client can prove it already has the current representation. It is orthogonal to whether the
+server itself re-computed that representation.
+
+They compose. §4 is built entirely around the observation that Event Caching's existing
+machinery — a pre-execution cache lookup, sitting in memory with the full rendered body, before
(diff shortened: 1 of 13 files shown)


lmajano:
Thanks for the review — all six findings were real and are fixed in 3e504d9:

- **`If-None-Match` matching**: now handles `*`, comma-separated lists, and always compares weakly per RFC 7232 §2.3.2/§3.2 (was exact-string-only against a single tag).
- **`If-Modified-Since` ignored when `If-None-Match` present**: per RFC 7232 §3.3 - `lastModified()` now checks for this before evaluating the date.
- **`toHTTPDate()` timezone**: now converts local→UTC via `dateConvert()` before formatting, so the trailing "GMT" is actually accurate. Fixed the RFC 7231 test to feed in the local equivalent of the UTC example so the assertion stays stable across CI timezones, per your suggestion.
- **Cache-hit replay duplicating/under-implementing matching logic**: `Bootstrap.cfc`'s cache-hit path now calls `event.etag()`/`event.lastModified()` directly instead of a separate hand-rolled comparison, so the automatic and manual paths share one implementation and can't drift apart again. This also means Last-Modified now actually short-circuits with a 304 on a cache hit, not just replays the header.
- **`etagWeak` not persisted on cache write**: now stored on the cache entry and read back correctly on hit.
- **Stale spec doc / PR description**: `docs/specs/http-caching.md`'s Status line and the PR description's mention of a `this.httpCaching.enabled` switch (removed earlier in review) are corrected.

New test coverage for the wildcard/list/weak matching and the If-None-Match precedence rule is in the same commit.

---

PR #686, opened 2026-08-23
fix: support full null runtimes

elpete:
## Summary

ColdBox relied on missing struct keys evaluating like null in a number of lazy-initialization and optional-input paths. Full-null runtimes instead omit those keys, so direct access raised undefined-element errors while Quick's expanded full-null matrix initialized ColdBox and ran application requests.

This change makes those paths explicitly key-aware across framework lifecycle, LogBox, CacheBox, WireBox, rendering, routing, REST, async tasks, and testing helpers. It also makes HTTP Last-Modified comparison use the RFC 1123 GMT instant so Adobe and Lucee agree.

Supports coldbox-modules/quick#154 and coldbox-modules/quick#312.

## Regression coverage

The regressions exercise public APIs rather than implementation details:

- Env.getJavaSystem() lazy initialization
- Util.getMixerUtil() lazy initialization
- LogBoxConfig.loadDataDSL() with a minimal public DSL
- CacheBoxConfig.loadDataDSL() with a minimal public DSL
- RequestContext.lastModified() HTTP-date behavior
- a dedicated Adobe 2025 full-null HTTP request that executes the same public APIs in a full-null application

Before the source fix, the Adobe endpoint failed with Element JAVASYSTEM is undefined. Additional full-suite failures exposed the other missing-key paths and were retained as fixes until the complete matrix was clean.

## Validation

- Adobe 2025 full null: 1,329 passed, 0 failed, 0 errored, 7 skipped
- Lucee 6 full null: 1,304 passed, 0 failed, 0 errored, 2 skipped
- Lucee 6 normal null mode: 1,304 passed, 0 failed, 0 errored, 2 skipped
- cfformat checks passed for changed sources and regression fixtures
- git diff --check passed

Local ColdBox validation used the TestBox full-null corrections proposed in Ortus-Solutions/TestBox#200. The self-contained Adobe endpoint added here does not depend on that PR and is included directly in CI.


lmajano:
Just for the record. This sucks on those engines. Because isNull() is supposed to work for any dereferenced target. Their implementation is a pain I never wanted to address

PR #688, opened 2026-08-25
fix: convert CacheBox minute timeouts in BoxLangProvider

dougcain:
`BoxLangProvider` does not convert CacheBox's minute-based timeouts before handing them to BoxLang's cache, which reads a bare number as **seconds**. Every timeout therefore expires sixty times too soon — a region moved from `CacheBoxProvider` to `BoxLangProvider` keeps a 10 minute object for 10 seconds.

`LuceeProvider` and `CFProvider` both convert. `BoxLangProvider` is the only one that doesn't.

## Evidence

CacheBox's contract is minutes — `ConcurrentStore.set` documents `@timeout Timeout in minutes`, and `CacheBoxProvider`, `LuceeProvider` and `CFProvider` all treat it that way.

Measured against a Redis-backed BoxLang region:

| call | resulting TTL |
|---|---|
| `set( key, value, 10 )` | **10 seconds** |
| `set( key, value, createTimespan( 0, 0, 10, 0 ) )` | 600 seconds |

The existing conversions this mirrors:

- `LuceeProvider.setQuiet` — `arguments.timeout = createTimespan( 0, 0, arguments.timeout, 0 )`
- `CFProvider.setQuiet` — same

## The change

A private `toTimespan()` applied to `timeout` and `lastAccessTimeout` in `set()`, `setQuiet()` and `getOrSet()`.

An already-converted value passes through untouched, and **cannot be detected with `isNumeric()`** — BoxLang's `createTimespan()` returns a `java.time.Duration`, and `isNumeric()` answers `true` for one, so a Duration would otherwise be fed back through `createTimespan()` as though it were a minute count. `isInstanceOf( ..., "java.time.Duration" )` is used instead.

Everything else passes through unchanged — `""` and `0` both mean "no timeout" to CacheBox — so the only behaviour that changes is the case that was wrong.

## Why this wasn't caught

`tests/specs/cache/providers/BoxLangProviderTest.cfc` only ever passes timespans:

```cfc
cache.set( "test", testVal, createTimespan( 0, 0, 2, 0 ), createTimespan( 0, 0, 1, 0 ) );
assertEquals( 120, md.timeout );
```

No existing test passes a bare minute count, so the missing conversion is invisible. This PR adds `testSetWithMinuteTimeouts`, `testSetQuietWithMinuteTimeouts` and `testSetLeavesATimespanAlone`.

## How it surfaced

cbSSO caches outstanding SAML AuthnRequest ids for 10 minutes and rejects any response that does not match one. Its docs recommend pointing `samlRequestCacheName` at a distributed CacheBox cache when running behind a load balancer — which on BoxLang means `BoxLangProvider`. The window silently became 10 seconds, so every SSO login slower than that failed. See coldbox-modules/cbSSO#22 for the downstream write-up.


lmajano:
Jira ticket: [COLDBOX-1422](https://ortussolutions.atlassian.net/browse/COLDBOX-1422)

---

[COLDBOX-1422]: https://ortussolutions.atlassian.net/browse/COLDBOX-1422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #695, opened 2026-09-21
fix: module convention routing skips optional-action fallback when a module declares a mandatory-action route

lmajano:
# Description

`ModuleService.cfc` auto-injects a `"/:handler/:action?"` convention route for any module that doesn't already have one, so single-segment (handler-only) URLs resolve within that module's scope. The "does one already exist" check uses this regex:

```
"^/?\:handler/\:action\??/?$"
```

The `\??` makes the literal `?` character itself *optional in the match*, so a module's own **mandatory**-action route (`"/:handler/:action"`, no `?`) satisfies this check — the regex treats it as equivalent to the **optional**-action route it's meant to detect. When a module declares only the mandatory variant, ColdBox incorrectly assumes an equivalent convention route already exists and skips auto-injecting the one that actually handles handler-only URLs. Any single-segment request to that module then falls through the module boundary entirely and gets handled by whatever the parent application's own top-level router does with it.

## Real-world impact

This is user-visible in ContentBox: `contentbox-admin`'s `config/Router.cfc` declares `route( "/:handler/:action" ).end();` (mandatory action) as its convention fallback. Clicking a first-level admin link like `/cbadmin/entries` or `/cbadmin/pages` silently lands on the front-end site instead of the admin section, while deeper links (3+ segments, which match an explicit route) work fine. I bisected this against ContentBox's `development` branch running on ColdBox `development` (this repo's current HEAD) vs. ColdBox `8.1.0` (where it does not reproduce), and root-caused it to this check.

## Fix

Require the literal `?` in the detection regex (`\:action\?` instead of `\:action\??`), so only a route that actually provides handler-only matching capability suppresses the auto-injection.

I verified the fix against a live ContentBox instance (development branch) with **no ContentBox-side changes** — the admin routing regression is fully resolved.

## Test plan

- Added a regression test in `ModuleServiceTest.cfc` with a minimal fixture module (`tests/resources/test-module-conventions`) that mirrors `contentbox-admin`'s setup: it declares only the mandatory-action route. The test asserts a handler-only URL still resolves within the module via `RoutingService.findRoute()`, instead of returning an empty route.
- Confirmed the new test fails without the fix (reverted the regex change locally, test fails with the exact expected message) and passes with it.
- Full `tests/specs/web` suite: **468/468 passing** (467 pre-existing + 1 new), no regressions.

## Jira Issues

I don't have access to this project's Jira tracker to file an issue — happy to file one if pointed to the right project/component, or a maintainer can link this to an existing/new ticket.

## Type of change

- [x] Bug Fix

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes (468/468 in `tests/specs/web`)
- [ ] I have made corresponding changes to the documentation (no user-facing docs affected; this restores existing documented behavior)

https://claude.ai/code/session_01CoEgT3ggNEKM1nGjqB5Qd5

---
_Generated by [Claude Code](https://claude.ai/code/session_01CoEgT3ggNEKM1nGjqB5Qd5)_


lmajano:
## Adobe CF test failures on this PR are a pre-existing flake, not caused by this change

The `tests / Test Suites (adobe@2023, ...)` and `tests / Test Suites (adobe@2025, ...)` jobs (and their `TestBox Report` counterparts) intermittently fail with:

```
Error loading module routes as the module requested 'domainRoutingTest' is not loaded.
```

thrown from `Router.cfc:323` inside `addModuleRoutes()`, originating in `tests/specs/web/routing/RoutingServiceTest.cfc:305` — the **"preserves the request domain when resolving module routes"** spec. This test is unrelated to this PR's change (it exercises domain-scoped route matching, not the mandatory/optional-action convention detection this PR fixes).

**Evidence this is a pre-existing flake, not something this PR introduced:**
1. **Same commit, different outcome across duplicate concurrent runs**: for head commit `4e051681` (this PR), two workflow runs triggered on the same push disagreed — `tests / Test Suites (adobe@2023, ...)` **passed** in run `35650497834` and **failed** in run `35650492147`, on the byte-identical commit.
2. **Same historical pattern on `development` itself, long before this PR existed**: the "ColdBox Test Suites" workflow, manually dispatched repeatedly against the exact same `development` commit `64a3c81a` back in May, alternates `success`/`failure`/`success`/`failure` across consecutive runs with zero code changes between them.

I don't have permission to trigger a re-run here (`rerun_failed_jobs` returned a 403), and I don't have an existing fix to port — the flakiness looks like a test-isolation issue (likely parallel TestBox workers racing on the shared `modules` config-settings struct), which is a larger, unrelated architectural question I don't want to speculatively fix inside this PR's scope.

Every check that actually exercises this PR's change is green: `boxlang@1`, `boxlang-cfml@1`, `boxlang-cfml@be`, `lucee@5`, `lucee@6`, and source formatting all pass consistently across every run, including the new regression test added here.

https://claude.ai/code/session_01CoEgT3ggNEKM1nGjqB5Qd5

---

PR #700, opened 2026-09-23
feat(perf-harness): compare BE against 8.1 stable and 7.x latest, fix BoxLang compat

lmajano:
## Description

Extends `tests/perf-harness/` to compare bleeding-edge (8.2-dev) against both ColdBox 8.1 stable and 7.x latest, across all four supported engines (BoxLang, BoxLang CFML, Adobe CF 2025, Lucee 7), and fixes a string of bugs that were silently preventing the suite from actually running at all.

**New capability**
- Added a `seven-app` fixture (mirrors `stable-app`, installs ColdBox 7.x) so the suite can compare 3 versions instead of 2.
- Generalized `PerformanceSuite.cfc` from a hardcoded BE-vs-stable comparison to an N-version comparison (`be`, `stable`, `seven`), with delta columns/rows computed dynamically for whichever versions were actually tested.
- Added a `perf:seven` convenience script to `box.json`.

**Bug fixes uncovered while actually running the suite for the first time**
- `cfhttp()` as a bracket-syntax call directly inside a `try{}` block trips a parser bug on the CLI's bundled Lucee 5.4.8.2; `continue` without a trailing semicolon inside a nested `if`/`for` hit the same engine's parser (see `AGENTS.md`'s `continue;`/`break;` compatibility note). Ultimately replaced `cfhttp()` entirely with a plain `java.net.HttpURLConnection` call, since the engine's `result="varName"` attribute also silently failed to populate when called from a component method.
- A `log()` helper name collided with the built-in CFML `Log()` BIF, which this engine's compiler resolves in preference to a same-named local UDF; renamed to `logMsg()`.
- `.flag("force")` isn't a real CommandDSL method; corrected to `.flags("force")`.
- Markdown headings and a CSS color used unescaped single `#` characters, which CFML treats as the start of an interpolated expression.
- A `%%` copy-paste holdover rendered literally in delta/error percentages (CFML has no such escape); fixed to a single `%`.
- **BoxLang canonicalization bug**: BoxLang derives a component's canonical type name from its physical path relative to the webroot, not from a `this.mappings` alias. With the shared repo-root webroot, `stable-app`/`seven-app`'s nested `coldbox/` install caused every typed-argument check inside ColdBox's own source (e.g. `LogEvent`) to fail with a type mismatch, producing 100% request errors on both BoxLang engines. Fixed by adding `this.mappings["/"]` in each app's `Application.cfc`, pointing at the app's own folder — verified this doesn't regress Lucee 7 or Adobe CF 2025, which already worked.
- `precisionEvaluate()` isn't implemented on BoxLang; a shared perf-harness fixture (`ProductService.cfc`) used it, breaking the "Complex View" scenario on every BoxLang-based engine/version. Replaced with `round(x * 100) / 100`.

**Known remaining gap (not fixed here, out of scope for this PR):** ColdBox 7.x still fails on native BoxLang (not BoxLang CFML) — its `InterceptorService.cfc` assumes a component's `extends` metadata key is absent when there's no superclass, but BoxLang always includes it as an empty struct, so `.name` access throws. This is a ColdBox 7.x / BoxLang compatibility gap in vendored framework code, not something fixable from the perf-harness's `Application.cfc`.

Full 4-engine × 3-version run (50 iterations, cold start measured) confirms BE has no meaningful latency regression vs 8.1 or 7.x on any working engine/scenario combination.

## Jira Issues

No Jira issue tracked for this internal dev-tooling change.

## Type of change

- [x] Improvement

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — N/A, this is a standalone perf-testing harness (`tests/perf-harness/`) run manually via `box task run`, not part of the TestBox suite
- [x] New and existing unit tests pass locally with my changes — no framework code was touched, only `tests/perf-harness/` tooling and fixtures

https://claude.ai/code/session_01KiK6DMek9iJMuYk2PzcjPj

---
_Generated by [Claude Code](https://claude.ai/code/session_01KiK6DMek9iJMuYk2PzcjPj)_


lmajano:
CI has now failed twice on this exact commit (`ecb5a6b8c`), each time on a different, unrelated job:

- Push-triggered run: `tests / Test Suites (lucee@6, 21, false, 8)` — internal Lucee engine NPE (`Cannot invoke "lucee.runtime.Mapping.getPageSource(String)" because "mapping" is null`) while running the main TestBox runner, before any spec even executes.
- This PR's run: `TestBox Report adobe@2023-11` — one erroring spec, `tests/specs/ioc/BuilderTest.cfc:162: The feed package is not installed.`

Neither failure touches anything this PR changes (`tests/perf-harness/*` only — no WireBox, no IoC, no core framework or engine-mapping code). `development`'s own scheduled Daily Test Run at HEAD (`5afcc6bc`) also failed the same way on yet another engine (`adobe@2025`) today, confirming this is pre-existing environmental flakiness in CI, not something introduced here.

I don't have permission to trigger a re-run (`rerun_failed_jobs` returns a 403), so I can't clear these myself. Flagging for a maintainer to re-run the checks, or let me know if you'd like anything else from me here.

---

lmajano:
Another commit (`78ea75d9`, a JSON-only version pin in `tests/perf-harness/`), another unrelated CI flake: `TestBox Report adobe@2023-11` errored on `tests.specs.integration.EventCachingSpec: can do pdf cached events` — `Error: The document package is not installed.` (in `system/core/conversion/DataMarshaller.cfc`). Same class of issue as the `feed` package error reported earlier, and the `lucee@6` internal NPE that also recurred on this same commit — none in code this PR touches.

Still no permission to re-run checks myself. Continuing to watch the PR; will keep flagging new distinct failures here rather than re-running speculative pushes against them.

---

lmajano:
Another one, same root cause pattern: `adobe@2025` failed with `The zip package is not installed` (in `runner-logbox.cfm`) — a 4th distinct missing-optional-cfpm-package flake (`feed`, `document`, now `zip`), plus the recurring `lucee@6` NPE. This looks like Adobe's `cfpm install` step in CI is unreliable, unrelated to this PR's diff. Not re-commenting per-instance further unless something changes — the PR just needs a maintainer re-run when convenient.

---

PR #701, opened 2026-09-23
perf: close BE request-lifecycle regression vs 8.1 stable

lmajano:
## Description

Follow-up to the perf-harness work in #700, which surfaced a consistent throughput regression on ColdBox BE (8.2-dev) vs 8.1 stable across engines. This PR fixes five per-request hot-path overheads found by diffing v8.1.0 against `development` and validating each candidate with paired BE/stable throughput measurements before turning it into a real fix.

**Fixes**

- `HandlerService.getRouteCachingMetadata()` (route-level `.withCache()` support) was resolved twice per request — once from `getEventMetadataEntry()` (pre-dispatch cacheability check) and again from `getEventCachingMetadata()` (at dispatch) — for a route record that cannot change for the life of a request. Memoized in `request` scope so it's resolved once.
- The cache-key-suffix dynamic-vs-static check (`isClosure()`/`isCustomFunction()`/`isSimpleValue()`) was recomputed via function calls on every cache-metadata lookup. Precomputed once as a boolean field (`suffixIsDynamic`) when the dictionary entry is built instead.
- `BaseService.getLogger()` dropped a redundant `structKeyExists()` check ahead of `isNull()` — the declared `log` property already guarantees the variables-scope slot exists.
- `InterceptorState` (extends `EventPool`, a separate class hierarchy from `BaseService`, so it had independently reimplemented the same lazy-getter pattern) now resolves the logger eagerly in `init()` instead of lazily on every call.
- `InterceptorService.getLazyBuffer()` was allocating a new `InterceptorBuffer` component on every `announce()` call (10+ per request). Pooled per request with checkout/release, falling back to an unpooled instance for the `async`/`asyncAll` paths (whose buffer can outlive the `announce()` call on a background thread) and for a reentrant `announce()` call (e.g. `onException` triggered from within an interceptor), so a nested call can never corrupt a buffer still in flight further up the call stack.

## Validation

- Functional: hit all 5 perf-harness scenarios plus an invalid-event path (to force `onInvalidEvent`/`onException`, exercising `announce()` reentrancy) against live BoxLang and Lucee 7 servers — all correct, no errors.
- Full TestBox suite could not be run in this sandbox (needs a MySQL datasource and a Lucee ORM extension, neither available here) — this is an infrastructure gap unrelated to the change, flagging for CI to confirm.
- Performance: repeated paired BE/8.1-stable throughput runs (150 iterations, 20s throughput window) on boxlang, boxlang-cfml, and lucee-7 show the BE/stable gap narrowed or closed, though this sandbox has significant run-to-run noise (documented throughout the investigation in #700) so treat exact percentages as directional, not precise.

## Type of change

- [x] Performance improvement

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — these are internal micro-optimizations to existing hot paths with no behavior change; correctness is covered by the existing framework test suite (not runnable in this sandbox, see above) and manual functional validation
- [x] New and existing unit tests pass locally with my changes — unable to run the full suite in this sandbox (see Validation); no behavior change intended, only per-request memoization/pooling of existing computations

https://claude.ai/code/session_01KiK6DMek9iJMuYk2PzcjPj

---
_Generated by [Claude Code](https://claude.ai/code/session_01KiK6DMek9iJMuYk2PzcjPj)_


lmajano:
One `tests / Test Suites (adobe@2025, 21, false, 8)` run failed on this commit (7cdfc7d2) with:

```
tests/specs/web/services/ModuleServiceTest.cfc | Error loading module routes as the module requested 'test-module-conventions' is not loaded.
```

This isn't this PR's failure — the stack trace only touches `Router.cfc` and `ModuleService.cfc` (module route registration), neither of which this PR modifies (this PR only touches `HandlerService.cfc`, `BaseService.cfc`, `InterceptorService.cfc`, and `InterceptorState.cfc`). Two CI workflow runs fired for this push; the sibling run's identical `adobe@2025` job on the same commit passed cleanly (870/870), which serves as the confirming re-run. Treating this as a pre-existing flake, likely from the two duplicate concurrent runs contending over module-registration state.

---

